PR #5, opened 2022-10-25
Create GET Pipelines endpoint

mmorgantaylor:


service/src/main/java/bio/terra/pipelines/app/controller/TspsApiController.java
```diff
@@ -11,21 +12,21 @@
...
+  // -- Pipelines --
 
-  // -- Policy Attribute Objects --
   @Override
-  public ResponseEntity<Void> deletePao(UUID objectId) {
-    paoService.deletePao(objectId);
-    return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+  public ResponseEntity<List> getPipelines() {
+    List<Pipeline> pipelineList = pipelinesService.getPipelines();
+    //    ApiPipelinesGetResult result = pipelineList;
```
mmorgantaylor, line 27: need help here - skipped this and it worked, but probably need to add something like this back in

service/src/main/java/bio/terra/pipelines/config/StatusCheckConfiguration.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.pipelines.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "javatemplate.status-check")
```
ahaessly, line 5: should we replace "javatemplate"?

service/src/test/java/bio/terra/pipelines/BaseSpringBootTest.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.pipelines;
+
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@SpringBootTest
+@ActiveProfiles({"test", "human-readable-logging"})
+// @ContextConfiguration(classes = App.class)
```
ahaessly, line 8: I'm not sure if I commented this out or if you did. Just wanted to point it out in case it was me. I'm not sure what it exactly does.

ahaessly (approved):
2 things for you to check.

service/src/main/java/bio/terra/pipelines/config/StatusCheckConfiguration.java
mmorgantaylor: ah great catch, thank you!!

service/src/test/java/bio/terra/pipelines/BaseSpringBootTest.java
mmorgantaylor: ah right, I did - thanks, I'll remove it

PR #9, opened 2022-11-28
[TSPS-7] README updates for new developers

mmorgantaylor:
This PR adds information to the README to help a developer who's new to the project get started. This includes requirements for programs that must be installed on the dev's local machine, as well as instructions for getting the service running locally.

We also needed to update the docker image used to access vault to make it compatible to be run on M1 machines (specifically: use an ARM-compatible build), per a recommendation from DevOps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,11 @@ To run locally:
 
 This service is written in Java 17, and uses Postgres 13. 
 
+To run locally, you'll also need: 
+- jq - install with `brew install jq`
+- vault - see DSP's setup instructions [here](https://docs.google.com/document/d/11pZE-GqeZFeSOG0UpGg_xyTDQpgBRfr0MLxpxvvQgEw/edit#heading=h.1k9ij99wmle2)
+  - Note that for Step 7, "Create a GitHub Personal Access Token", you'll want to choose the "Tokens (classic)" option, not the fine-grained access token option.
+
 ### Tech stack
 
 - Java 17 temurin
--- a/README.md
+++ b/README.md
@@ -1,3 +1,34 @@
 # Terra Scientific Pipelines Service
 
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=DataBiosphere_terra-scientific-pipelines-service&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=DataBiosphere_terra-scientific-pipelines-service)
+
+## Overview
+
+Terra Scientific Pipelines Service, or teaspoons (tsps), facilitates running a number of defined scientific pipelines 
+on behalf of users that users can't run themselves in Terra. The most common reason for this is that the pipeline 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Development
+
+This codebase is in initial development.
+
+To run locally:
+1. Make sure you have the requirements (below) installed. We recommend IntelliJ as an IDE.
+2. Clone the repo
+3. Run the commands in `scripts/postgres-init.sql` in your local postgres instance. You will need to be authenticated to access Vault.
+4. Run `scripts/write-config.sh`
```
koncheto-broad, line 29: This step failed for several reasons when I tried to run it on my M1 Mac.  Two things needed to occur for it to pass successfully.

1. I needed to update the script locally to point to the docker image 

 A default run of write-config.sh will try to use the ARM image and fail with an obscure stack trace saying
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
runtime: failed to create new OS thread (have 2 already; errno=22)
fatal error: newosproc

runtime stack:
runtime.throw(0x18d8566, 0x9)
	/goroot/src/runtime/panic.go:596 +0x95
runtime.newosproc(0xc420024000, 0xc420034000)
	/goroot/src/runtime/os_linux.go:163 +0x18c
[...]
```
0.20.0 is an older docker image that does not have am ARM-compatible build (according to Mike Flinn in #dsp-devops-champions), and the dev or master images have those builds.  So I just changed line 148 of the file to reference
`broadinstitute/dsde-toolbox:dev`
instead of 
`broadinstitute/dsde-toolbox:consul-0.20.0`



2. I needed to install vault.  Vault was mentioned in passing in step 3, but it has to be explicitly installed and some setup has to be done.  The instructions for getting authenticated with vault can be found here: https://docs.google.com/document/d/11pZE-GqeZFeSOG0UpGg_xyTDQpgBRfr0MLxpxvvQgEw/edit#heading=h.trehvi2yr632

The one caveat worth mentioning is that Step 7 in the instructions, where they give instructions for creating a GitHub Personal Access Token now only applies to the ones listed under "Tokens (classic)."  I have not explored how to make an acceptable fine-grained token that works with this, so as long as they continue offering the "classic" tokens developers should use those.

koncheto-broad: oh, one last thing.  The script still fails if you don't have the program jq installed, which on a Mac is fixed as simply as running
`brew install jq`

But if you don't install it, the config script will fail.

koncheto-broad (commented):
Added a few notes from my experience getting my dev environment set up.

README.md
mmorgantaylor: thanks so much! I think I've resolved everything here - would you mind taking a look to confirm? 

(I had to delete my `config` directory to get the script to re-run, but it worked on my end with my M1.)

koncheto-broad (approved):
LGTM

PR #12, opened 2023-03-06
[TSPS-8] Validate job request (check for existing pipelineId)

mmorgantaylor:
This PR ensures that when a new job request is made, the `pipelineId` in the request does exist (in the pipelines db table). If the provided `pipelineId` does not exist, we throw a 404 Pipeline not found error and return it to the user.

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/db/PipelinesDao.java
+++ b/service/src/main/java/bio/terra/pipelines/db/PipelinesDao.java
@@ -1,12 +1,9 @@
 package bio.terra.pipelines.db;
 
-import bio.terra.common.exception.NotFoundException;
 import bio.terra.pipelines.app.configuration.TspsDatabaseConfiguration;
 import bio.terra.pipelines.service.model.Pipeline;
 import java.util.*;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.dao.EmptyResultDataAccessException;
-import org.springframework.dao.support.DataAccessUtils;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
@@ -34,13 +31,10 @@ public class PipelinesDao {
    * @param pipelineId
    * @return Pipeline object
    */
-  public Pipeline getPipeline(String pipelineId) {
-    DbPipeline dbPipeline =
-        getDbPipelineIfExists(pipelineId)
-            .orElseThrow(
-                () -> new NotFoundException(String.format("Pipeline %s not found.", pipelineId)));
+  public Boolean checkPipelineExists(String pipelineId) {
+    Integer dbPipelineCount = countDbPipeline(pipelineId);
 
-    return Pipeline.fromDb(dbPipeline);
+    return dbPipelineCount == 1;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


service/src/main/java/bio/terra/pipelines/db/PipelinesDao.java
```diff
@@ -17,11 +21,26 @@ public class PipelinesDao {
...
+   * Return the matching pipeline, given the pipelineId and version
+   *
+   * @param pipelineId
+   * @return Pipeline object
+   */
+  public Pipeline getPipeline(String pipelineId) {
+    DbPipeline dbPipeline =
+        getDbPipelineIfExists(pipelineId)
+            .orElseThrow(
+                () -> new NotFoundException(String.format("Pipeline %s not found.", pipelineId)));
```
mmorgantaylor, line 41: check to see whether the common library calls NotFoundException elsewhere

service/src/main/java/bio/terra/pipelines/db/PipelinesDao.java
```diff
@@ -17,11 +21,26 @@ public class PipelinesDao {
...
-  private final JdbcTemplate tpsJdbcTemplate;
+  private final NamedParameterJdbcTemplate jdbcTemplate;
 
   @Autowired
   public PipelinesDao(TspsDatabaseConfiguration tspsDatabaseConfiguration) {
-    this.tpsJdbcTemplate = new JdbcTemplate(tspsDatabaseConfiguration.getDataSource());
+    this.jdbcTemplate = new NamedParameterJdbcTemplate(tspsDatabaseConfiguration.getDataSource());
+  }
+
+  /**
```
mmorgantaylor, line 31: rather than a getPipeline method, could use a true checkPipelineExists or getPipelineCount method that does essentially a SELECT COUNT(pipeline_id) so that it doesn't pull data out that we're not using, and we can make the error checking more clear at a high level

service/src/main/java/bio/terra/pipelines/db/PipelinesDao.java
mmorgantaylor: as discussed on Slack: terra-common-lib doesn't have other calls of NotFoundException, but other repos (e.g. WSM) do create custom classes for each [Object]NotFoundException, so we're opting to follow the apparent DSP convention and create a PipelineNotFoundException class

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -24,4 +25,13 @@ public List<Pipeline> getPipelines() {
...
     return pipelinesDao.getPipelines();
   }
+
+  public void validatePipeline(String pipelineId) {
+    logger.info("Validate pipeline");
+
+    Boolean pipelineExists = pipelinesDao.checkPipelineExists(pipelineId);
+    if (Boolean.FALSE.equals(pipelineExists)) {
+      throw new PipelineNotFoundException(String.format("Pipeline %s not found.", pipelineId));
+    }
```
kilonzi, line 35:
```suggestion
    if (!pipelineExists) {
      throw new PipelineNotFoundException(String.format("Pipeline %s not found.", pipelineId));
    }
```
wondering if that is more readable

kilonzi (approved):
Just one minor suggestion

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: ugh, it totally is more readable, and that's what I had originally, but SonarCloud complained that it wouldn't handle a `null` properly and gave this suggestion instead. 🙃

kilonzi: I would trust SonarCloud or have it return false by default

mmorgantaylor: good point. I'm trying out having the validatePipeline() method return a (non-nullable) `boolean` rather than `Boolean`. we'll see if that makes SonarCloud happy with `!pipelineExists`...

mmorgantaylor: SonarCloud is pleased with our efforts! thanks so much for the suggestion.

PR #15, opened 2023-03-08
[TSPS-11] format empty timeCompleted to null in API response

mmorgantaylor:
This PR processes the Optional object returned by timeCompleted in the getJob(s) responses into a null rather than a string "Optional.empty"

I edited the objectMapper bean in App.java to remove a line that removes null values from API responses; we prefer to include `"timeCompleted": null` in the response rather than remove the "timeCompleted" field entirely.


ahaessly (approved):
👍 looks great!

PR #16, opened 2023-03-09
[TSPS-12] store and return timestamps with UTC timezone

mmorgantaylor:
this PR ensures that timestamps are stored in UTC and returned by APIs in UTC. we do this by using `Instant` rather than `Timestamp`, as the former enforces UTC.


koncheto-broad (approved):
LGTM

PR #19, opened 2023-03-17
[TSPS-29] Testing coverage for JobsService

koncheto-broad:
Adding tests for the parts of JobsService that contain actual busines logic.  Also making the UUID generation protected, so I can override it using a spy.

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/service/JobsServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/JobsServiceTest.java
@@ -1,0 +1,85 @@
+package bio.terra.pipelines.service;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.*;
+
+import bio.terra.pipelines.db.JobsDao;
+import bio.terra.pipelines.db.exception.DuplicateObjectException;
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/pipelines/service/JobsServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  // We'll need these to configure the dao to return selectively good or bad values
+  private final UUID testGoodUUID = UUID.randomUUID();
+  private final UUID testBadUUID = UUID.randomUUID();
+
+  @BeforeEach
+  void initMocks() {
+    // dao throws on job containing bad id and returns good uuid on job containing good uuid
+    when(jobsDao.createJob(argThat((Job j) -> j.getJobId() == testBadUUID)))
+        .thenThrow(DuplicateObjectException.class);
```
mmorgantaylor, line 36: `jobsDao.createJob()` now returns `null` if a duplicate is found - I think we need to update this logic and the `testCreateJob_unsuccessfulWriteDaoThrows()` test

koncheto-broad: I believe it doesn't behave that way yet (relative to the current state of merged code in main).  Is TSPS-19 ready to be reviewed and merged yet?  If not, we can update the tests when it is.  Otherwise, I'll have to check in tests that are currently broken.

PR #20, opened 2023-03-20
[TSPS-31] refactor pipelineId check to be called from controller layer

mmorgantaylor:
This PR:
- changes the PipelinesService method from `validatePipeline()` (throwing an exception if pipelineId not found) to `pipelineExists()` (returning a boolean)
- moves the validation step to call `pipelineExists()` from JobsService to JobsApiController
- updates PipelinesService tests for `validatePipeline()`
- DOES NOT include testing of validation in JobsApiController - this should be added in TSPS-27

Note - we probably want to invoke validatePipeline() in the other pipelines API requests - made TSPS-33 to do that

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -77,6 +77,8 @@ paths:
           $ref: '#/components/responses/BadRequest'
         '403':
           $ref: '#/components/responses/PermissionDenied'
+        '404':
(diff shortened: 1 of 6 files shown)


koncheto-broad (approved):
LGTM!

PR #21, opened 2023-03-21
[TSPS-32] convert timeCreated Instant to timestamp before writing to db

mmorgantaylor:
this PR converts the Instant class `timeCreated` to a Timestamp before writing to database (in JobsDao)

note this PR does NOT include tests! those should be added for this functionality in TSPS-30.


koncheto-broad (approved):
This definitely fixes the problem for now, and we should merge it.  We might want to discuss later whether or not we handle this inside of the Job object, and in general how we handle the fact that Instants give us the _behavior_ we want but Timestamps are required as arguments for the SQL layer.

PR #24, opened 2023-04-05
[TSPS-38] Update README with instructions for connecting to database and deployment

mmorgantaylor:


koncheto-broad (approved):
Look forward to trying this out myself!

PR #25, opened 2023-04-10
[Tsps-27] Testing coverage for JobsApiController

koncheto-broad:
Expanding test coverage for JobsApiController.

- Uncommenting negative case for getJob
- Expanding testGetMessageOk case to test content of response object
- Testing positive and negative cases for createJob
- Testing positive cases for getJobs
- Adding json conversion method to MockMvcUtils

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/controller/JobsApiControllerTest.java
+++ b/service/src/test/java/bio/terra/pipelines/controller/JobsApiControllerTest.java
@@ -42,7 +42,7 @@ import org.springframework.test.web.servlet.MvcResult;
 
 @ContextConfiguration(classes = {JobsApiController.class, GlobalExceptionHandler.class})
 @WebMvcTest()
-class JobsControllerTest {
+class JobsApiControllerTest {
   @MockBean JobsService jobsServiceMock;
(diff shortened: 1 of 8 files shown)


service/src/test/java/bio/terra/pipelines/controller/JobsApiControllerTest.java
```diff
@@ -0,0 +1,197 @@
...
+        mockMvc
+            .perform(get(String.format("/api/jobs/v1alpha1/%s/%s", pipelineId, jobIdOkDone)))
+            .andExpect(status().isOk())
+            .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+            .andReturn();
+
+    ApiGetJobResponse response =
+        new ObjectMapper()
+            .readValue(result.getResponse().getContentAsString(), ApiGetJobResponse.class);
+    // you could compare other fields here too beyond the id, if wanted
```
mmorgantaylor, line 104: I would think we would want to assert that all the expected fields are present in the response. any reason not to do this?

mmorgantaylor (approved):
looks good, thanks! one question but otherwise 👍

PR #27, opened 2023-05-01
Updating db types

koncheto-broad:
I got rid of the `text` type in the jobs and pipelines tables.  I used an upper bound of 50 for the main ids, an upper bound of 200 for display names and versions, and a generous upper bound of 1000 for freeform text fields like status and description.  This should give us plenty of room in the fields, and we can always tweak the values later if they cause problems (or trim them down, if we see that they are too generous.

### DIFF
--- a/service/src/main/resources/db/changesets/20230427.yaml
+++ b/service/src/main/resources/db/changesets/20230427.yaml
@@ -1,4 +1,4 @@
-# Add jobs table
+# Switch text types to varchar whenever possible
 databaseChangeLog:
   - changeSet:
       id: modifying text column types to varchar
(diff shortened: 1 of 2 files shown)


service/src/main/resources/db/changesets/20230427.yaml
```diff
@@ -0,0 +1,32 @@
+# Add jobs table
```
mmorgantaylor, line 1: super minor, but could you update this comment or remove it (since there's the changeSet.id field already with a description)?

mmorgantaylor (approved):
one very minor comment request, otherwise looks good, thanks!

service/src/main/resources/db/changesets/20230427.yaml
koncheto-broad: oops!  I'll get rid of that comment

PR #28, opened 2023-05-19
TSPS-41 add PipelineInputs field to PostJobRequestBody

jsotobroad:
I think this should do what we want, I definitely took a few wrong paths 😬.  I did think about using the [AnyValue type](https://swagger.io/docs/specification/data-models/data-types/#any) option but seemed a little too "blobby" for this.  I couldn't find any relevant tests to update since its just an empty object atm so not really anything to verify but let me know if I missed something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ To run locally, you'll also need:
 - vault - see DSP's setup instructions [here](https://docs.google.com/document/d/11pZE-GqeZFeSOG0UpGg_xyTDQpgBRfr0MLxpxvvQgEw/edit#heading=h.1k9ij99wmle2)
   - Note that for Step 7, "Create a GitHub Personal Access Token", you'll want to choose
     the "Tokens (classic)" option, not the fine-grained access token option.
+- Java 17 - can be installed manually or through IntelliJ which will do it for you when importing the project
+- Postgres 13 - multiple solutions here as long as you have a postgres instance running on localhost:5432 the local app will connect appropriately
+  - Postgres.app https://postgresapp.com/
+  - Brew https://formulae.brew.sh/formula/postgresql@13
 
 ### Tech stack
 
@@ -45,12 +49,19 @@ To run locally, you'll also need:
 ### Local development
 
 To run locally:
-1. Make sure you have the requirements (below) installed. We recommend IntelliJ as an IDE.
-2. Clone the repo
+1. Make sure you have the requirements installed from above. We recommend IntelliJ as an IDE.
+2. Clone the repo (if you see broken inputs build the project to get the generated sources)
 3. Run the commands in `scripts/postgres-init.sql` in your local postgres instance. You will need to be authenticated to access Vault.
 4. Run `scripts/write-config.sh`
 5. Run `./gradlew bootRun`
 6. Navigate to [http://localhost:8080/#](http://localhost:8080/#)
+
+### Running Tests/Linter Locally
+- Testing
+  - Run `./gradlew service:test` to run tests
+- Linting
+  - Run `./gradlew spotlessCheck` to run linter checks 
+  - Run `./gradlew :service:spotlessApply` to apply fix any issues the linter finds
 
 ### Running SonarQube locally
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -32,6 +32,10 @@ To run locally, you'll also need:
 - vault - see DSP's setup instructions [here](https://docs.google.com/document/d/11pZE-GqeZFeSOG0UpGg_xyTDQpgBRfr0MLxpxvvQgEw/edit#heading=h.1k9ij99wmle2)
   - Note that for Step 7, "Create a GitHub Personal Access Token", you'll want to choose
     the "Tokens (classic)" option, not the fine-grained access token option.
+- Java 17 - can be installed manually or through IntelliJ which will do it for you when importing the project
+- Postgres 13 - multiple solutions here as long as you have a postgres instance running on localhost:5432 the local app will connect appropriately
+  - Postgress.app https://postgresapp.com/
```
mmorgantaylor, line 37: minor: typo on Postgress.app (extra s)

service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
```diff
@@ -62,18 +62,20 @@ public ResponseEntity<ApiPostJobResponse> createJob(
     final SamUser userRequest = getAuthenticatedInfo();
     String userId = userRequest.getSubjectId();
     String pipelineVersion = body.getPipelineVersion();
+    Object pipelineInputs = body.getPipelineInputs();
```
mmorgantaylor, line 65: I think you probably want to make changes to the tests in JobsApiControllerTest.java - since the new "pipeline inputs" input should be required

mmorgantaylor (changes requested):
looks good overall, i think you do want to make the new input a required field in the payload, which I think should cause existing tests to fail as they are now, so those will need to be updated

minor: in future it'd be great to have separate PRs for different tasks (like the readme update vs the work for TSPS-41) but it's not a big deal here since the readme update is so minor

common/openapi.yml
```diff
@@ -224,14 +224,16 @@ components:
...
     # Please keep alphabetized
-
     PostJobRequestBody:
       description: |
         Object containing the user-provided information about a job request.
       type: object
       properties:
         pipelineVersion:
           $ref: "#/components/schemas/PipelineVersion"
+        pipelineInputs:
```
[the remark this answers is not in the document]
jsotobroad: makes sense, also added pipelineVersion to required since it is currently not nullable in the database

jsotobroad:
> minor: in future it'd be great to have separate PRs for different tasks (like the readme update vs the work for TSPS-41) but it's not a big deal here since the readme update is so minor

totally, i had them separate but then i wanted to add extra changes while working on this code so i just put it all together 😛 but agreed!

common/openapi.yml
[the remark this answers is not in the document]
mmorgantaylor: good call, thanks for fixing that!

service/src/test/java/bio/terra/pipelines/testutils/MockMvcUtils.java
```diff
@@ -33,6 +33,8 @@ public static MockHttpServletRequestBuilder addJsonContentType(
   public static String convertToJsonString(Object obj) throws JsonProcessingException {
     ObjectMapper mapper = new ObjectMapper();
     mapper.configure(SerializationFeature.WRAP_ROOT_VALUE, false);
+    // needed because we currently allow amorphous objects currently as pipelineInputs
+    mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
jsotobroad, line 37: i dont love that I did this but I think once we are validating inputs to maybe non amorphous classes we can get rid of it.  Open to any suggestions though

mmorgantaylor: hm. would it help to make it a key-value-pair structure, like a HashMap? (dictionary in Python)

mmorgantaylor (approved):

PR #29, opened 2023-05-24
TSPS-42 store pipeline_inputs in the jobs table

jsotobroad:
Still a bit of weirdness with the testing of the currently amorphous object as the pipeline inputs, having to use linkedmaps in some places as that is the object that naturally comes back from the jobs table query.  The actual jobsdaotest is nice though.

I stored things as a string since I dont think there is any reason we would want to be querying the inputs from the db directly so we dont need jsonb (but definitely let me know if this isnt the case anymore.

some screenshots showing how its stored now to make sure thats what we expect.
Screenshot 2023-05-24 at 6 11 29 PM

Screenshot 2023-05-24 at 6 02 04 PM

Screenshot 2023-05-24 at 6 09 41 PM

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/db/JobsDao.java
+++ b/service/src/main/java/bio/terra/pipelines/db/JobsDao.java
@@ -1,7 +1,6 @@
 package bio.terra.pipelines.db;
 
-import static java.sql.Types.TIMESTAMP;
-import static java.sql.Types.VARCHAR;
+import static java.sql.Types.*;
 
 import bio.terra.common.db.ReadTransaction;
 import bio.terra.common.db.WriteTransaction;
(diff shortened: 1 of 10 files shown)


service/src/main/java/bio/terra/pipelines/db/JobsDao.java
```diff
@@ -133,7 +136,7 @@ public List<Job> getJobs(String userId, String pipelineId) {
   private Optional<DbJob> getDbJobIfExists(String userId, String pipelineId, String jobId) {
     final String sql =
         """
-                        SELECT job_id, user_id, pipeline_id, pipeline_version, time_submitted, time_completed, status
+                        SELECT job_id, user_id, pipeline_id, pipeline_version, time_submitted, time_completed, status, pipeline_inputs
```
jsotobroad, line 139: Was thinking we might want to separate the pipeline_inputs into its own table so we don't have to keep returning that data (which may be large in some cases) when we just want to return the job information like we currently do for the 2 jobs GET endpoints.  Something simple like

|job_id|pipeline_inputs|
|---|---|
|UUID|Text|

PR #31, opened 2023-05-31
TSPS-42 save pipeline inputs to a new table

jsotobroad:
created a new table, pipeline_inputs, that is used to save the pipeline inputs provided through the api.  for now storing inputs as a string as we dont expect to need to parse the inputs directly through the database which would make jsonb the more correct solution.

screenshots showing requests and subsequent database store

Screenshot 2023-05-31 at 12 17 53 PM

Screenshot 2023-05-31 at 12 18 59 PM

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
@@ -80,7 +80,8 @@ public class JobsApiController implements JobsApi {
     // TODO assuming we will write outputs back to source workspace, we will need to check user
     // permissions for write access to the workspace - explore interceptors
 
-    UUID createdJobUuid = jobsService.createJob(userId, pipelineId, pipelineVersion);
+    UUID createdJobUuid =
+        jobsService.createJob(userId, pipelineId, pipelineVersion, pipelineInputs);
     if (createdJobUuid == null) {
       logger.error("New {} pipeline job creation failed.", pipelineId);
       throw new ApiException("An internal error occurred.");
@@ -116,13 +117,13 @@ public class JobsApiController implements JobsApi {
 
   static ApiGetJobResponse jobToApi(Job job) {
     return new ApiGetJobResponse()
-        .jobId(job.getJobId().toString())
-        .userId(job.getUserId())
-        .pipelineId(job.getPipelineId())
-        .pipelineVersion(job.getPipelineVersion())
-        .timeSubmitted(job.getTimeSubmitted().toString())
-        .timeCompleted(job.getTimeCompleted().map(Instant::toString).orElse(null))
-        .status(job.getStatus());
+        .jobId(job.jobId().toString())
+        .userId(job.userId())
+        .pipelineId(job.pipelineId())
+        .pipelineVersion(job.pipelineVersion())
+        .timeSubmitted(job.timeSubmitted().toString())
+        .timeCompleted(job.timeCompleted().map(Instant::toString).orElse(null))
+        .status(job.status());
   }
 
   static ApiGetJobsResponse jobsToApi(List<Job> jobList) {
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
@@ -39,9 +39,9 @@ public class PipelinesApiController implements PipelinesApi {
     for (Pipeline pipeline : pipelineList) {
       var apiPipeline =
           new ApiPipeline()
-              .pipelineId(pipeline.getPipelineId())
-              .displayName(pipeline.getDisplayName())
-              .description(pipeline.getDescription());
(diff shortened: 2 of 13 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
```diff
@@ -116,13 +117,13 @@ public ResponseEntity<ApiGetJobsResponse> getJobs(@PathVariable("pipelineId") St
...
   static ApiGetJobResponse jobToApi(Job job) {
     return new ApiGetJobResponse()
-        .jobId(job.getJobId().toString())
-        .userId(job.getUserId())
-        .pipelineId(job.getPipelineId())
-        .pipelineVersion(job.getPipelineVersion())
-        .timeSubmitted(job.getTimeSubmitted().toString())
-        .timeCompleted(job.getTimeCompleted().map(Instant::toString).orElse(null))
-        .status(job.getStatus());
+        .jobId(job.jobId().toString())
```
jsotobroad, line 120: part of converting to a record

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -39,9 +39,9 @@ static ApiPipelinesGetResult pipelinesToApi(List<Pipeline> pipelineList) {
     for (Pipeline pipeline : pipelineList) {
       var apiPipeline =
           new ApiPipeline()
-              .pipelineId(pipeline.getPipelineId())
-              .displayName(pipeline.getDisplayName())
-              .description(pipeline.getDescription());
+              .pipelineId(pipeline.pipelineId())
```
jsotobroad, line 42: record conversion

service/src/main/java/bio/terra/pipelines/db/JobsDao.java
```diff
@@ -57,26 +63,30 @@ public JobsDao(TspsDatabaseConfiguration tspsDatabaseConfiguration) {
...
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue(JOB_ID_PARAM, jobUuid.toString(), VARCHAR)
-            .addValue(USER_ID_PARAM, job.getUserId(), VARCHAR)
-            .addValue(PIPELINE_ID_PARAM, job.getPipelineId(), VARCHAR)
-            .addValue(PIPELINE_VERSION_PARAM, job.getPipelineVersion(), VARCHAR)
-            .addValue(TIME_SUBMITTED_PARAM, Timestamp.from(job.getTimeSubmitted()), TIMESTAMP)
-            .addValue(STATUS_PARAM, job.getStatus(), VARCHAR);
+            .addValue(USER_ID_PARAM, job.userId(), VARCHAR)
+            .addValue(PIPELINE_ID_PARAM, job.pipelineId(), VARCHAR)
```
jsotobroad, line 79: record conversion

service/src/main/java/bio/terra/pipelines/service/model/Job.java
```diff
@@ -5,59 +5,14 @@
 import java.util.Optional;
 import java.util.UUID;
 
-public class Job {
```
jsotobroad, line 8: converted this class to a record

service/src/test/java/bio/terra/pipelines/db/PipelinesDaoTest.java
```diff
@@ -18,9 +18,9 @@ void testGetPipelines() {
     assertEquals(retrievedPipelines.size(), nTotalPipelines);
 
     for (Pipeline pipeline : retrievedPipelines) {
-      assertNotNull(pipeline.getPipelineId());
-      assertNotNull(pipeline.getDisplayName());
-      assertNotNull(pipeline.getDescription());
+      assertNotNull(pipeline.pipelineId());
+      assertNotNull(pipeline.displayName());
```
jsotobroad, line 22: record change

jsotobroad (commented):
hmm i converted the Job and Pipeline classes to records per a sonarqube suggestion, it changed more files that i had realized but I tried commenting all the places that this changed in this pr vs the parts that actually have to do with the functionality around saving pipeline inputs to the new table.  if its too annoying i can split out those changes.

koncheto-broad:
I've looked over the PR and it looks fine.  I know we were going to have a conversation with some of the other teams about using different orm-like db frameworks (we think they're great, but there's value to being in sync with what the other teams are doing if/when engineers get shuffled between projects), and should that hold for converting all db-focused model objects to records?  I'm not sure I've seen that used heavily in other terra projects, and it would be nice to have at least had a brief discussion with them to make sure nobody has any objections (it can be as quick as asking "SonarCube is pushing for immutable classes to be records--do we have any compelling reason internally to avoid doing so?").

I learn on using records and whichever frameworks make sense for us, as long as they aren't esoteric.

jsotobroad:
closed as this is pre data JPA being used in the project.  gonna rewrite

PR #35, opened 2023-07-18
TSPS-42 add pipeline_inputs table 

jsotobroad:
create a table to save the pipeline inputs passed in by the user when submitting a job to our service

### DIFF
--- a/service/src/main/resources/db/changesets/20230711.yaml
+++ b/service/src/main/resources/db/changesets/20230711.yaml
@@ -11,13 +11,13 @@ databaseChangeLog:
             columns:
               - column:
                   name: job_id
-                  type: UUID
+                  type: int
                   constraints:
                     nullable: false
                     primaryKey: true
                     foreignKeyName: pipeline_inputs_job_id_FK
                     referencedTableName: jobs
-                    referencedColumnNames: job_id
+                    referencedColumnNames: id
               - column:
                   name: inputs
                   type: text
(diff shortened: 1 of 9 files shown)


service/src/main/resources/db/changesets/20230711.yaml
```diff
@@ -0,0 +1,25 @@
...
+            remarks: |
+              There is one row in the pipeline_inputs table per job submitted to the service.
+            columns:
+              - column:
+                  name: job_id
+                  type: UUID
+                  constraints:
+                    nullable: false
+                    primaryKey: true
+                    foreignKeyName: pipeline_inputs_job_id_FK
```
jsotobroad, line 18: I wasn't sure if it mad more sense for the FK to be `jobs.job_id` or `jobs.id` and i went with `jobs.job_id`.  Curious on your thoughts on this

koncheto-broad: It's not a huge deal, but I'd normally just stuck with jobs.id.  Then, in any other tables that use it as a fk, I'd make THOSE other_table.job_id.  But again, it ultimately doesn't matter much.

koncheto-broad (approved):
LGTM.  Left non-essential feedback on the open question you had, and glad to see that you cleaned up a little of the cruft that I left behind in the JobsService test class

jsotobroad:
@koncheto-broad if you could give this a quick once over after the changes.  I changed the FK to be jobs.id like we talked about

PR #36, opened 2023-07-31
TSPS-67 added notebook interacting with WDS and CBAS

jsotobroad:
Notebook used to help with de-risking WDS and CBAS.  Has all of the functions we would need to interact with both of those apps. We have some open questions around what actually needs to be provided when submitting a submission in terms of workflow_inputs and workflow_outputs but that will be settled outside of this notebook's code.  

I'm happy to do a walkthrough before this is merged as well


koncheto-broad (approved):
LGTM

PR #37, opened 2023-08-07
TSPS-69 update file copy testing notebook 

jsotobroad:
Added the commands to copy from the "control" workspace to the "user" workspace as well.  Added some more comments to the existing notebook.  Removed the TSPS function calls


koncheto-broad (approved):
LGTM!  Good to have this in the repo

PR #42, opened 2023-11-20
TSPS-100 add logic to use service account credentials when talking to leonardo

jsotobroad:
### Description 

Changing our usage of the leonardo client to use the google default credentials available instead of user credentials.  These can be supplied manually when developing and will be provided as part of the deployment in live environments

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-100

### DIFF
--- a/scripts/write-config.sh
+++ b/scripts/write-config.sh
@@ -91,7 +91,7 @@ case $target in
         ;;
 
     local)
-        # for local development we will use the QA environemnt stuff for now to mimic our BEEs
+        # for local development we will use the QA environment stuff for now to mimic our BEEs
         fcenv=qa
         ;;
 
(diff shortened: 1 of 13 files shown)


koncheto-broad (approved):
Not 100% sure what the ImputationService class will ultimately do vs the flights that actually pull together the disparate pieces necessary to use our service end to end, but let's see where this goes!

scripts/write-config.sh
```diff
@@ -93,28 +91,16 @@ case $target in
         ;;
 
     local)
-        k8senv=integration
-        # TODO: probably not right, but makes the
-        namespace=wsmtest
-        fcenv=dev
+        # for local development we will use the QA environemnt stuff for now to mimic our BEEs
```
mmorgantaylor, line 94: nit: typo "environment"

mmorgantaylor (approved):
super nice! thanks for updating the README and organizing the dependencies.

PR #46, opened 2023-12-03
TSPS-102 add instrumentation of prometheus metrics

jsotobroad:
### Description 

adding the instrumentation to connect metrics to prometheus by adding in actuator and exporting a count metric for pipelines run.  We'll want to have a discussion for what other product level metrics we want to track but just putting this one in as a test for now. This will need to go with a infrastructure helmchart change in order to get metrics to show up in a grafana dashboard.

screenshot of metric being tracked in my local server
Screenshot 2023-12-03 at 1 03 02 PM

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-102


mmorgantaylor (approved):
looks awesome, thanks!

PR #47, opened 2023-12-12
TSPS-118 add CreateJob flight

mmorgantaylor:
### Description 

This PR removes the toy example GetPipeline flight and implements the start of a CreateJob flight. The CreateJob flight currently has just two steps: one that sets the status to SUBMITTED and logs a timeSubmitted in the working map, and one that writes all the job info to the TSPS database. 

Not included in this PR but tickets for future work:
- TSPS-123: Update API endpoints to bring us in line with Terra recommendations
- TSPS-124: Add an integration test for CreateJob
- TSPS-125: Specify the precise Stairway Steps for the CreateJob endpoint
- TSPS-128: Make Stairway tests run independently

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-118


service/src/main/java/bio/terra/pipelines/common/utils/JobStatusEnum.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.pipelines.common.utils;
+
+public enum JobStatusEnum {
```
mmorgantaylor, line 3: I imagine in future adding more descriptive statuses like `COPYING INPUT DATA` and `RUNNING IMPUTATION PIPELINE` but we can discuss.

likely will be renamed to something more descriptive like `TspsJobStatusEnum` when we do the async API refactoring

service/src/main/java/bio/terra/pipelines/stairway/PlaceholderSetStatusToSubmittedStep.java
```diff
@@ -0,0 +1,39 @@
...
+  /* This is a placeholder step that only sets the status in the working map;
+  it will be replaced with real steps in future PRs */
+  public PlaceholderSetStatusToSubmittedStep(JobsService jobsService) {
+    this.jobsService = jobsService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) throws InterruptedException {
+
+    // to add later: submit the workflow to CBAS
```
mmorgantaylor, line 23: or we might decide that "timeSubmitted" means when the USER submitted the job, in which case this step wouldn't involve submitting to CBAS. in any case this will get changed a bit as we figure out how we want our Steps to work

service/src/main/java/bio/terra/pipelines/common/utils/JobStatusEnum.java
jsotobroad: Yea so i think these should be statuses of what step the pipeline might be on.  Maybe having a parent class of `CommenJobStatusEnum` and then a subclass of `ImputationPipelineJobStatusEnum`.

service/src/main/java/bio/terra/pipelines/service/JobsService.java
```diff
@@ -23,56 +27,60 @@ public class JobsService {
...
     logger.info("Create new {} version {} job for user {}", pipelineId, pipelineVersion, userId);
 
-    // placeholder for actually doing something; for now we're just writing the info to the database
-    //        JobBuilder createJob = jobService ...
-
-    String status = "SUBMITTED";
-
-    return writeJobToDb(
-        userId, pipelineId, pipelineVersion, timeSubmitted, status, pipelineInputs, 1);
+    StairwayJobBuilder stairwayJobBuilder =
```
jsotobroad, line 59: is this where we would want to distinguish what stairwayjob we're created based on pipelineId or would that happen.  like would the `.flightClass()` value be dependent on the piplineId?

service/src/main/java/bio/terra/pipelines/service/JobsService.java
```diff
@@ -23,56 +27,60 @@ public class JobsService {
...
+            .pipelineVersion(pipelineVersion)
+            .submittingUserId(userId)
+            .pipelineInputs(pipelineInputs);
+
+    return stairwayJobBuilder.submit();
   }
 
-  protected UUID createJobId() {
-    return UUID.randomUUID();
+  protected String createJobId() {
```
jsotobroad, line 72: do we need this anymore?  doesnt stairway have its own way of generating job ids?

service/src/main/java/bio/terra/pipelines/stairway/CreateJobFlight.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override
+  public void addStep(Step step, RetryRule retryRule) {
+    super.addStep(step, retryRule);
+  }
+
+  public CreateJobFlight(FlightMap inputParameters, Object beanBag) {
+    super(inputParameters, beanBag);
+    final FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(beanBag);
+
+    FlightUtils.validateRequiredEntries(
```
jsotobroad, line 23: awesome

service/src/main/java/bio/terra/pipelines/stairway/PlaceholderSetStatusToSubmittedStep.java
jsotobroad: yea ths jobs table should just be the life of the tsps workflow/job.  so we could have a `submitted` and `modified` columns.

service/src/main/java/bio/terra/pipelines/service/JobsService.java
```diff
@@ -83,48 +91,33 @@ protected UUID writeJobToDb(
...
+            String.format("Duplicate jobId %s found", job.getJobId()));
       }
       throw e;
     }
 
     return job;
   }
 
-  private Instant getCurrentTimestamp() {
+  public Instant getCurrentTimestamp() {
```
jsotobroad, line 120: hmm i think we can use postgres itself to set these timestamps for us when rows are created in the database.  This is a mysql example but something like [this](https://stackoverflow.com/questions/19230489/default-value-on-update-liquibase)

service/src/main/java/bio/terra/pipelines/service/JobsService.java
jsotobroad: oh interesting, so if the flight fails to initialize, then there will be nothing written to the tsps database.  I feel like maybe this should happen before we launch the flight?  We should chat about what makes sense here.

service/src/main/java/bio/terra/pipelines/stairway/WriteJobToDbStep.java
```diff
@@ -0,0 +1,59 @@
...
+        CreateJobFlightMapKeys.PIPELINE_ID,
+        CreateJobFlightMapKeys.PIPELINE_VERSION,
+        CreateJobFlightMapKeys.SUBMITTING_USER_ID);
+
+    var workingMap = flightContext.getWorkingMap();
+    FlightUtils.validateRequiredEntries(workingMap, CreateJobFlightMapKeys.STATUS);
+
+    UUID writtenJobUUID =
+        jobsService.writeJobToDb(
+            UUID.fromString(flightContext.getFlightId()),
```
jsotobroad, line 39: hmm generally you dont want to have to keep going from strings to UUIDs and vice versa in different parts of the code.  Is there any way to streamline this so its always a UUID wherever the value is?

service/src/main/java/bio/terra/pipelines/service/JobsService.java
mmorgantaylor: yeah exactly, I think we probably want to write some stuff to the db right away. but I think that those decisions should be mapped out in [TSPS-125](https://broadworkbench.atlassian.net/browse/TSPS-125) and don't need to be solidified for this PR necessarily

service/src/main/java/bio/terra/pipelines/stairway/WriteJobToDbStep.java
mmorgantaylor: I made some changes, but this particular one I'm having trouble getting around, since FlightContext is a Stairway class, and .getFlight() returns a String. I agree that it's gross to keep going back and forth... open to suggestions here!

service/src/main/java/bio/terra/pipelines/service/JobsService.java
mmorgantaylor: great call. made TSPS-129 to do this

service/src/main/java/bio/terra/pipelines/service/JobsService.java
jsotobroad: 👍

service/src/main/java/bio/terra/pipelines/stairway/WriteJobToDbStep.java
jsotobroad: i think the changes you've made are alot better.  All of the transitiions between types are at teh stairway interaction level where I think it makes the most sense for it to be

service/src/main/java/bio/terra/pipelines/app/configuration/internal/StairwayDatabaseConfiguration.java
```diff
@@ -7,7 +7,7 @@
 
 @Configuration
 @EnableConfigurationProperties
-@ConfigurationProperties(prefix = "pipelines.stairway-database")
+@ConfigurationProperties(prefix = "spring.stairway-database")
```
jsotobroad, line 10: ~would it make sense to put this under `"stairway.stairway-database"` instead?~ nvm i think this makes sense with the other things we have under "spring"

service/src/test/java/bio/terra/pipelines/testutils/MockMvcUtils.java
```diff
@@ -47,12 +49,15 @@ public static String convertToJsonString(Object obj) throws JsonProcessingExcept
           new BearerToken(UUID.randomUUID().toString()));
 
   // Pipelines test constants
-  public static final String TEST_PIPELINE_ID_1 = "test-pipeline-id-1";
-  public static final String TEST_PIPELINE_VERSION_1 = "1.0.0";
-  public static final String TEST_PIPELINE_NAME_1 = "Test Pipeline Name One";
-  public static final String TEST_PIPELINE_DESCRIPTION_1 = "Test Pipeline Description One";
-  public static final String TEST_PIPELINE_ID_2 = "test-pipeline-id-2";
-  public static final String TEST_PIPELINE_VERSION_2 = "1.0.0";
+  public static final String TEST_PIPELINE_ID_1 =
```
jsotobroad, line 52: nice grouping of commonly used test utils.  Was thinking maybe we want to rename the file or split the mvc stuff from like general `TestUtil` stuff but not something we have to decide now, just ToL

service/src/test/resources/init-create-stairway-db.sql
```diff
@@ -0,0 +1 @@
+CREATE DATABASE tsps_stairway_db;
```
jsotobroad, line 1: \n (yes im a pain sorry)

jsotobroad (approved):
some small comments but looks great.  I think you have a good plan to move to from this with all the tickets you created 👍

koncheto-broad (approved):
This, in combination with the new tickets, looks awesome Morgan!

service/src/test/resources/init-create-stairway-db.sql
mmorgantaylor: haha no, thank you for catching these!

service/src/test/java/bio/terra/pipelines/testutils/MockMvcUtils.java
mmorgantaylor: i like it. easy enough to refactor now.

PR #48, opened 2024-01-09
TSPS-123 Refactor to follow DSP's Async API best practices

mmorgantaylor:
### Description 

This PR brings our service closer to being in line with DSP's Asynchronous API best practices [doc](https://docs.google.com/document/d/1PTd4xvmV9xnEkWaIgFc6d3VUyJwwymQTFdhpKrdsUKw/edit#heading=h.ol9mx3vfhjjj). Remaining work for this goal that is not included in this PR is encapsulated in the following tickets:
- TSPS-135: Implement GET imputation/result/{id} endpoint
- TSPS-136: Require the caller to provide the job UUID
- TSPS-150: Write tests for JobApiUtils

The plan of changes can be found in [this doc](https://docs.google.com/spreadsheets/d/19DRc0HvBjXYHNk_EfhmVXd40e5R1V73bwybqM8uP6b0/edit#gid=1471653780). 

The changes in this PR are as follows:

API endpoint and controller changes: 
- Create the following API endpoints:
  - `POST /api/pipelines/v1alpha1/{pipelineId}` to create a new pipeline of type {pipelineId} e.g. imputation. The response to this request now contains a `JobControl` object (specified by the best practices doc) that contains the jobId, rather than just the jobId.
  - `GET /api/pipelines/v1alpha1/{pipelineId}/jobs` to retrieve all pipeline-specific jobs requested by user. 
  - `GET /api/job/v1alpha1/jobs` to retrieve all jobs (across pipelines) requested by user
  - `GET /api/job/v1alpha1/jobs/{jobId}` to retrieve a specified job
- Note that the creation of a new job has moved from `JobsApiController` to `PipelinesApiController`, which checks which pipeline is requested (currently the only possible one is imputation) and then calls the appropriate service (ImputationService) to create the job.

Database changes: 
- The TSPS database's `jobs` table has been renamed to `imputation_jobs` (and future pipelines will get their own tables), and fields in this database that are captured in Stairway's database are removed (timestamps, status, and pipelineId). Currently, `imputation_jobs` stores the internal id, the jobId for the flight, the pipeline version, and the userId of the submitting user.
- Note that the `pipeline_inputs` table remains as-is and is not imputation-specific. We may make the inputs table pipeline-specific in future but this was not needed in this PR.

Stairway-related changes: 
- Stairway classes refactored e.g. `StairwayJobService` and `StairwayJobBuilder` are now `JobService` and `JobBuilder`
- Stairway JobBuilder now requires userId and pipelineId for all flights. It also requires a jobId (flightId) input, though the work to make the user supply this rather than the service is still to be completed in TSPS-136.
- Stairway JobService now performs job lookups, both for getting all jobs and for getting specific jobs. It always checks that the jobs returned were submitted by the requesting user.
- Add `JobApiUtils` class that contains useful methods for converting Stairway job objects (EnumeratedJobs, FlightStates, etc) to API response objects. Note that the `mapFlightStatusToApi` method groups the finer-grained Stairway statuses into three user-facing statuses for a job: `RUNNING`, `SUCCEEDED`, `FAILED`.

Other:
- Create `PipelinesEnum` to encompass validated pipelines. Note that there is a test in `PipelinesServiceTest` to ensure that every pipeline in `PipelinesEnum` has a corresponding row in the pipelines repository pipelines table.  

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-123


common/openapi.yml
```diff
@@ -23,126 +23,167 @@ paths:
...
           $ref: '#/components/responses/PermissionDenied'
-        '404':
+        404:
           $ref: '#/components/responses/NotFound'
-        '500':
+        500:
           $ref: '#/components/responses/ServerError'
     post:
-      summary: Create new job request
+      summary: Create new pipeline job
```
jsotobroad, line 81: just to make sure im on the right track, eventually this will require a jobid to be passed in, in which case we will also need to add a 202 reponse code when we do?

common/openapi.yml
```diff
@@ -23,126 +23,167 @@ paths:
...
+        200:
+            description: Success
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/GetJobsResponse'
+        403:
           $ref: '#/components/responses/PermissionDenied'
-        '404':
+        404:
```
jsotobroad, line 147: shouldnt need 403 or 404 here if we're just returning the list of jobs

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -58,4 +97,82 @@ static ApiPipeline pipelineToApi(Pipeline pipelineInfo) {
         .displayName(pipelineInfo.getDisplayName())
         .description(pipelineInfo.getDescription());
   }
+
+  // Pipelines jobs
```
jsotobroad, line 101: probably worth expanding on this comment since its probably the most integral endpoint we have

service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
```diff
@@ -61,93 +49,23 @@ private SamUser getAuthenticatedInfo() {
...
-  @Override
-  public ResponseEntity<ApiGetJobResponse> getJob(
-      @PathVariable("pipelineId") String pipelineId, @PathVariable("jobId") UUID jobId) {
-    final SamUser userRequest = getAuthenticatedInfo();
-    String userId = userRequest.getSubjectId();
-    Job job = jobsService.getJob(userId, pipelineId, jobId);
-    ApiGetJobResponse result = jobToApi(job);
-
+    logger.info("Retrieving jobId {} for userId {}", jobId, userId);
+    FlightState flightState = stairwayJobService.retrieveJob(jobId, userId);
```
jsotobroad, line 56: how are we making sure that we are correctly handling 403s (dont have permission to see job id being asked for) and 404s (job id doesnt exist)

service/src/main/java/bio/terra/pipelines/common/utils/PipelinesEnum.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.pipelines.common.utils;
+
+public enum PipelinesEnum {
+  IMPUTATION("imputation");
```
jsotobroad, line 4: we should probably make a ticket to clean out the toy pipelines in the pipelines table.

service/src/main/java/bio/terra/pipelines/dependencies/stairway/StairwayJobMapKeys.java
```diff
@@ -19,9 +22,4 @@ public enum StairwayJobMapKeys {
   public String getKeyName() {
     return keyName;
   }
-
-  public static boolean isRequiredKey(String keyName) {
```
jsotobroad, line 23: is this something that isnt useful to have?  I can imagine a place where we want to get a list of required keys.  Was it not being used before?

service/src/main/java/bio/terra/pipelines/dependencies/stairway/StairwayJobService.java
```diff
@@ -216,12 +220,23 @@ public Stairway getStairway() {
...
 
   @Traced
-  public FlightState retrieveJob(UUID jobId) {
+  public FlightState retrieveJob(UUID jobId, String userId) {
     try {
-      return stairwayComponent.get().getFlightState(jobId.toString());
+      FlightState result = stairwayComponent.get().getFlightState(jobId.toString());
+      // Note: after implementing TSPS-134, we can filter by flightId in enumerateJobs and remove
+      // the following check
+      if (!userId.equals(
```
jsotobroad, line 228: so i wonder if this scenario shoudl return a job not found or a "forbidden" exception.  Also at some point we will be doing these checks at the workspace level as opposed to the user level.  Is that something we want to do now or do it in another round of refactors

service/src/main/resources/db/changesets/20240104.yaml
```diff
@@ -0,0 +1,20 @@
...
+            tableName: imputation_jobs
+            columns:
+            - column:
+                name: pipeline_id
+            - column:
+                name: time_submitted
+            - column:
+                name: time_completed
+            - column:
+                name: status
```
jsotobroad, line 20: nice! we can clean up the pipelines table here if we want.

jsotobroad (commented):
were there any tests that you werent able to recapitulate with all of the refactoring you did or any parts of the code you wanted to test that felt overly difficult to test?  The coverage looks great but was just curious.

common/openapi.yml
mmorgantaylor: that's right - TSPS-136 covers requiring the jobid in the payload here. good catch about the 202 - that ought to be part of this PR, will add it.

common/openapi.yml
mmorgantaylor: good catch, thank you!

service/src/main/java/bio/terra/pipelines/dependencies/stairway/StairwayJobService.java
mmorgantaylor: re: checks at workspace level, I'd say make that change in the future if we do it, since that's only applicable if users are submitting data from a workspace, versus uploading the data in the no-BP model. or am i missing something?

service/src/main/resources/db/changesets/20240104.yaml
mmorgantaylor: good call, done!

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
mmorgantaylor: added descriptions to each controller method below - is this what you had in mind, or is there still missing context?

service/src/main/java/bio/terra/pipelines/common/utils/PipelinesEnum.java
mmorgantaylor: cleaned up in this PR, thanks for the reminder

service/src/main/java/bio/terra/pipelines/dependencies/stairway/StairwayJobMapKeys.java
mmorgantaylor: ugh, agreed, but the place I wanted to use it (when setting up a new flight) is written in a way that wasn't obvious how to use this. but I'm going to take another look, because this is a much clearer way to specify the required keys than what we have.

service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
mmorgantaylor: discussed in Slack - had been overly cautious with returning 404s instead of 403s, but this isn't necessary. should be returning 403s properly now.

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+      throw new MissingRequiredFieldException(
+          "Missing required field for flight construction: flightClass");
+    }
+
+    if (jobId == null) {
+      throw new MissingRequiredFieldException(
+          "Missing required field for flight construction: jobId");
+    }
+
+    for (JobMapKeys key : JobMapKeys.values()) {
```
jsotobroad, line 76: It might be a bit more efficient if we change `isRequiredKey` to be collection of values.  Then we can loop through that collection of values and check against `jobParameterMap` to see if they exist.

Also usually with validations like this you'd want to gather all failures at the same time as opposed to returning the first failure you see.  I think this is pretty much a check to be internally consistent so probably not a big deal but when we start validing user inputs and stuff we'll want to do that instead.

service/src/main/java/bio/terra/pipelines/service/ImputationService.java
```diff
@@ -1,40 +1,140 @@
...
+            .addParameter(JobMapKeys.DESCRIPTION.getKeyName(), description)
+            .addParameter(RunImputationJobFlightMapKeys.PIPELINE_VERSION, pipelineVersion)
+            .addParameter(RunImputationJobFlightMapKeys.PIPELINE_INPUTS, pipelineInputs);
+
+    return jobBuilder.submit();
+  }
+
+  // only used in tests
+  @VisibleForTesting
+  public ImputationJob getImputationJob(UUID jobId, String userId) {
```
jsotobroad, line 97: do we need this function and the one below it if theyre only used for testing?  cant we just call the imputationJobsRepository in the tests directly?

service/src/main/java/bio/terra/pipelines/service/ImputationService.java
```diff
@@ -1,40 +1,140 @@
...
+   * successful.
+   *
+   * @param userId
+   * @param pipelineVersion
+   * @return String jobId
+   *     <p>Note that the information in the requested job will grow over time, along with the
+   *     following related classes:
+   * @see ImputationJob
+   */
+  @Transactional
```
jsotobroad, line 76: does this need to be transactional.  I'm assuming this is just doing one write to the stairway database

service/src/main/java/bio/terra/pipelines/service/ImputationService.java
mmorgantaylor: :facepalm: sure can, thanks!

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobBuilder.java
mmorgantaylor: great points, will do

service/src/main/java/bio/terra/pipelines/service/ImputationService.java
mmorgantaylor: hmm at this time that's true. in future we might be doing more that might merit its being transactional, but happy to take this out for now

jsotobroad (approved):

PR #52, opened 2024-01-30
TSPS-136 Require a JobControl object containing a job id when creating a new job

mmorgantaylor:
### Description 

Per DSP's Async API best practices, the caller must provide a job id contained in a JobControl object. This PR implements that requirement and validates that the string the user has provided is a uuid.

Some other minor edits:
- removed our `MissingRequiredFieldsException` class and use TCL's `MissingRequiredFieldException` instead
- removed the user's email address (PII) from the log message
- changed the base class for StairwayDatabaseConfigurationTest from BaseTest to BaseEmbeddedDbTest, which allows that test to pass for me locally.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-136


common/openapi.yml
```diff
@@ -361,12 +361,14 @@ components:
...
         Object containing the user-provided information about a job request.
       type: object
-      required: [ pipelineVersion, pipelineInputs ]
+      required: [ jobControl, pipelineVersion, pipelineInputs ]
       properties:
         description:
           description: >-
             User-provided description of the job request.
           type: string
+        jobControl:
```
jsotobroad, line 370: just curious, what is the benefit of this schema being an object vs a string like pipelineVersion?

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -56,6 +58,30 @@ public ResponseEntity<ApiErrorReport> validationExceptionHandler(Exception ex) {
...
+    Map<String, String> errors = new HashMap<>();
+    ex.getBindingResult()
+        .getAllErrors()
+        .forEach(
+            error -> {
+              String fieldName = ((FieldError) error).getField();
+              String errorMessage = error.getDefaultMessage();
+              errors.put(fieldName, errorMessage);
+            });
+    String validationErrorMessage = "Request could not be parsed or was invalid: " + errors;
```
jsotobroad, line 77: i would probaby put at least a `\n` between each error so it can be a bit easier to read (if only slightly)

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -56,6 +58,30 @@ public ResponseEntity<ApiErrorReport> validationExceptionHandler(Exception ex) {
...
+    logger.debug("MethodArgumentNotValid exception caught by global exception handler", ex);
+    // For security reasons, we generally don't want to include the user's invalid (and potentially
+    // malicious) input in the error response, which also means we don't include the full exception.
+    // Instead, we return a generic error message about input validation.
+    Map<String, String> errors = new HashMap<>();
+    ex.getBindingResult()
+        .getAllErrors()
+        .forEach(
+            error -> {
+              String fieldName = ((FieldError) error).getField();
```
jsotobroad, line 73: do we have to do the cast here?  surprised thats necessary if we know what the exception we're starting with is.

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -119,17 +120,20 @@ public ResponseEntity<ApiCreateJobResponse> createJob(
       @PathVariable("pipelineId") String pipelineId, @RequestBody ApiCreateJobRequestBody body) {
     final SamUser userRequest = getAuthenticatedInfo();
     String userId = userRequest.getSubjectId();
+    String jobIdString = body.getJobControl().getId();
+    UUID jobId = validateJobId(jobIdString);
```
jsotobroad, line 124: what do we think abotu using the openapi yaml to validate this for us.  something like what is found in  https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#support-for-x-www-form-urlencoded-request-bodies

service/src/main/java/bio/terra/pipelines/service/ImputationService.java
```diff
@@ -77,13 +77,17 @@ public class ImputationService {
...
+      String description,
+      String pipelineVersion,
+      Object pipelineInputs) {
     logger.info("Create new imputation version {} job for user {}", pipelineVersion, userId);
 
     JobBuilder jobBuilder =
         jobService
             .newJob()
-            .jobId(createJobId())
+            .jobId(jobId)
```
jsotobroad, line 90: woooo! things are comin together

jsotobroad (commented):
curiosities and some questions

common/openapi.yml
mmorgantaylor: it's the spec. it's an object that contains currently just the "id" field, but in future it could additionally contain things like notification preferences. does this answer your question?

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
mmorgantaylor: nice, I didn't see the "format: uuid" option. that throws a different error that we have to handle differently in the GlobalExceptionHandler. I've implemented it... see what you think.

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
mmorgantaylor: we seemingly do... `.getAllErrors1 returns a List<ObjectError>, and ObjectError doesn't have a `.getField` method. let me know if I'm missing something though!

mmorgantaylor: nevermind! there's a `.getFieldErrors` method!

common/openapi.yml
jsotobroad: 👍

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
jsotobroad: woohoo

jsotobroad (approved):

PR #54, opened 2024-02-05
TSPS-135 add the get result endpoint

mmorgantaylor:
### Description 

- Add GET {pipelineName}/result/{jobId} endpoint
- This endpoint returns a CreatedJobResponse object that always includes a JobReport and depending on status returns an ErrorReport (if the job failed) or a PipelineOutput (if the job succeeded)
- changed the PipelineOutput format to String (this will likely be the path to the result workspace, but TBD)
- Add functionality to generate and return the ResultURL in the JobReport; the GET jobs/{jobid} endpoint now returns a JobReport that includes this URL

e.g. response to a createJob request is now
```
{
  "jobReport": {
    "id": "faf85f64-5717-4562-b3fc-2c963f66afa6",
    "description": "string",
    "status": "RUNNING",
    "statusCode": 202,
    "submitted": "2024-02-12T17:08:32.537386Z",
    "resultURL": "http://localhost:8080/api/pipelines/v1alpha1/imputation_minimac4/result/faf85f64-5717-4562-b3fc-2c963f66afa6"
  }
}
```

GET job/{jobid} response in a BEE: (note the resultURL)
```
{
  "id": "aaaabbbb-5717-4562-b3fc-2c963f66afa6",
  "description": "string",
  "status": "SUCCEEDED",
  "statusCode": 200,
  "submitted": "2024-02-13T15:28:43.782580Z",
  "completed": "2024-02-13T15:28:44.469910Z",
  "resultURL": "https://tsps.morgan.bee.envs-terra.bio/api/pipelines/v1alpha1/imputation_minimac4/result/aaaabbbb-5717-4562-b3fc-2c963f66afa6"
}
```

NOTE that currently our steps do not set the RESPONSE parameter, so the result endpoint if you run this doesn't return a PipelineOutput. but we do test that, when we do set a RESPONSE value in the working map, that value is returned as the PipelineOutput.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-135

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -124,6 +124,53 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
+  /api/pipelines/v1alpha1/{pipelineName}/result/{jobId}:
+    parameters:
+        - $ref: '#/components/parameters/PipelineName'
+        - $ref: '#/components/parameters/JobId'
+    get:
+      operationId: getPipelineJobResult
+      tags: [ pipelines ]
+      responses:
+        200:
+          description: Job is complete (succeeded or failed)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreateJobResponse'
+        202:
+          description: Job is running
+          headers:
+            Retry-After:
+              description: >-
+                optional - estimated seconds to wait before polling again. This allows a server to offer a hint as to when the job might be complete.
+              schema:
+                type: integer
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreateJobResponse'
+        400:
+          description: Bad request - invalid id, id not for a createJobRequest job
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        403:
+          description: No permission to see job
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Not found - job id does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/ServerError'
+
   /api/job/v1alpha1/jobs:
     get:
       tags: [ jobs ]
@@ -182,11 +229,10 @@ paths:
 components:
 
   parameters:
-
     JobId:
       name: jobId
       in: path
-      description: A UUID string identifier to used to identify a job in the service
+      description: A UUID identifier for a pipelines job
       required: true
       schema:
         type: string
@@ -340,10 +386,7 @@ components:
       required: [ id ]
       properties:
         id:
-          description: >-
-            Required unique identifier (UUID) for the job.
-          type: string
-          format: uuid
+          $ref: '#/components/schemas/Id'
         # In the future, notification configuration will also be part of JobControl.
 
     JobResult:
@@ -385,6 +428,8 @@ components:
           $ref: '#/components/schemas/JobReport'
         errorReport:
           $ref: '#/components/schemas/ErrorReport'
+        pipelineOutput:
+          $ref: '#/components/schemas/PipelineJobOutput'
 
     GetPipelinesResult:
       type: array
@@ -408,11 +453,11 @@ components:
           items:
             $ref: '#/components/schemas/JobReport'
 
-    JobId:
+    Id:
       description: |
-        Unique identifier (UUID) for a job request.
+        Required unique identifier (UUID) for a pipelines job.
       type: string
-      format: string
+      format: uuid
 
     JobStatus:
       description: |
@@ -462,6 +507,11 @@ components:
         The identifier string for the Pipeline.
       type: string
       format: string
+
+    PipelineJobOutput:
+      description: |
+          The output of a successful pipeline job.
+      type: object
 
     PipelineVersion:
       description: |
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -511,7 +511,7 @@ components:
     PipelineJobOutput:
       description: |
           The output of a successful pipeline job.
-      type: object
+      type: string
 
     PipelineVersion:
       description: |
--- a/service/src/main/java/bio/terra/pipelines/app/configuration/external/IngressConfiguration.java
+++ b/service/src/main/java/bio/terra/pipelines/app/configuration/external/IngressConfiguration.java
@@ -1,0 +1,22 @@
+package bio.terra.pipelines.app.configuration.external;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "pipelines.ingress")
+public class IngressConfiguration {
+
+  /** Fully-qualified domain name. The base URL this instance can be accessed at. */
+  private String domainName;
+
+  public String getDomainName() {
+    return domainName;
+  }
+
+  public void setDomainName(String domainName) {
+    this.domainName = domainName;
+  }
+}
--- a/service/src/main/java/bio/terra/pipelines/app/controller/JobApiUtils.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/JobApiUtils.java
@@ -2,10 +2,7 @@ package bio.terra.pipelines.app.controller;
 
 import bio.terra.common.exception.ErrorReportException;
 import bio.terra.pipelines.app.configuration.external.IngressConfiguration;
-import bio.terra.pipelines.common.utils.PipelinesEnum;
 import bio.terra.pipelines.dependencies.stairway.JobMapKeys;
-import bio.terra.pipelines.dependencies.stairway.JobService;
-import bio.terra.pipelines.dependencies.stairway.exception.InternalStairwayException;
 import bio.terra.pipelines.dependencies.stairway.exception.InvalidResultStateException;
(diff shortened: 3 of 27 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/JobApiUtils.java
```diff
@@ -121,13 +134,97 @@ public static ApiErrorReport buildApiErrorReport(Exception exception) {
...
+      return new AsyncJobResult<T>()
+          .jobReport(jobReport)
+          .result(resultOrException.getResult())
+          .errorReport(errorReport);
+    } catch (StairwayException stairwayEx) {
+      throw new InternalStairwayException(stairwayEx);
+    }
+  }
+
+  private String resultUrlFromFlightState(FlightState flightState) {
```
jsotobroad, line 179: so it looks like 2 of these functions are non static and the rest (4?) are static.  Is there any meaningful way we can split them up into two separate files?  If not i would probably leave the static ones as static.  Theres enough of them that its meaningful to have them be "correct" i think.

I wonder if the two non static methods could be moved into JobService?

service/src/main/java/bio/terra/pipelines/app/configuration/external/IngressConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "pipelines.ingress")
+public class IngressConfiguration {
+
+  /** Fully-qualified domain name. The base URL this instance can be accessed at. */
+  private String domainName;
```
jsotobroad, line 13: kinda funny this was already in our application.yml w/o us using it haha.

just to double check, does our helm-file overwrite this value when deploying to different environments?

common/openapi.yml
```diff
@@ -463,6 +508,11 @@ components:
       type: string
       format: string
 
+    PipelineJobOutput:
```
jsotobroad, line 511: im assuming this will get expanded as we figure out what things we want to return to the user upon a completed pipeline job?

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -170,11 +176,30 @@ public ResponseEntity<ApiGetJobsResponse> getPipelineJobs(
...
+    PipelinesEnum validatedPipelineId = validatePipelineName(pipelineName);
+
+    JobApiUtils.AsyncJobResult<String> jobResult =
+        jobApiUtils.retrieveAsyncJobResult(jobId, userId, validatedPipelineId, String.class, null);
+
+    ApiCreateJobResponse response =
+        new ApiCreateJobResponse()
+            .jobReport(jobResult.getJobReport())
+            .errorReport(jobResult.getApiErrorReport())
+            .pipelineOutput(jobResult.getResult()); // this is null unless the job succeeded
```
jsotobroad, line 198: does our flight return anything as a `pipelineOutput` when it is successful?

common/openapi.yml
mmorgantaylor: yes, exactly right

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
mmorgantaylor: our flight does not currently set anything to the RESPONSE field in the working map, so nothing is currently returned here, no.

service/src/main/java/bio/terra/pipelines/app/configuration/external/IngressConfiguration.java
jsotobroad: i think its actually more appropriate to have it called `tsps.ingress` so i would update the application.yml + the prefix here instead of the helm chart.

Also I dont think we're using the helm chart environment variable anywhere in our application.yml so we'll have to add that as well (no matter what we do with the naming)

common/openapi.yml
```diff
@@ -124,6 +124,53 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Not found - job id does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
```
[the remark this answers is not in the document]
mmorgantaylor: if the job itself had an internal error, this call will return a 200 (job completed, failed) and the 500 will be reflected in the ErrorReport. i think the 500 here means there was an internal error with the call itself, so we wouldn't want to return an ErrorReport reflecting the flight. does that make sense?

common/openapi.yml
```diff
@@ -408,11 +453,11 @@ components:
           items:
             $ref: '#/components/schemas/JobReport'
 
-    JobId:
```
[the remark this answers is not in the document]
mmorgantaylor: Ah, no, its reference got removed in my previous PR (once we properly return a JobControl with `id` field as the response to creating a new job, rather than just the jobId) and I failed to remove this. you can see that here: https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/48/files#diff-e862a1565882d38762f5b513f134435a5f220d06a8374341f5200de8a61ff131L264 (old row 264 in openapi.yml)

service/src/main/java/bio/terra/pipelines/app/configuration/external/IngressConfiguration.java
mmorgantaylor: added the direct reference (though the spring magic did work without it)

service/src/test/java/bio/terra/pipelines/configuration/external/IngressConfigurationTest.java
```diff
@@ -0,0 +1,17 @@
...
+import bio.terra.pipelines.testutils.BaseEmbeddedDbTest;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+class IngressConfigurationTest extends BaseEmbeddedDbTest {
+  @Autowired IngressConfiguration ingressConfiguration;
+
+  @Test
+  void verifyIngressConfiguration() {
+    assertNotNull(ingressConfiguration.getDomainName());
```
jsotobroad, line 15: you could also put a value in the application-test.yml and test against that value here if you want

service/src/test/java/bio/terra/pipelines/controller/JobApiUtilsTest.java
```diff
@@ -1,32 +1,53 @@
...
 import org.springframework.http.HttpStatus;
 
-class JobApiUtilsTest {
+class JobApiUtilsTest extends BaseEmbeddedDbTest {
+
+  @InjectMocks private JobApiUtils jobApiUtils;
+  @Mock private JobService jobService;
+  @Mock private IngressConfiguration ingressConfiguration;
+
+  // we mock IngressConfiguration.getDomainName() to return "localhost", which maps to "http://"
```
jsotobroad, line 37: you probably dont have to mock this if you add it to the application-test.yml.  You should be able to `@Autowired` it for any test class taht subclassed BaseEmbeddedDbTest.

At least im pretty sure this is possible, if you've tried it already and it didnt work then ignore :)

service/src/test/java/bio/terra/pipelines/dependencies/stairway/JobServiceMockTest.java
```diff
@@ -186,10 +191,11 @@ void retrieveJobResult_interrupted() throws InterruptedException {
   void retrieveJob_stairwayException() throws InterruptedException {
```
jsotobroad, line 191: I just noticed these test method names being both camel and snake case.  We should take a pass at these and make them consistent (camel case preferred)

jsotobroad (commented):
super nitpicky comments, we can chat about some of them and you can ignore others :)

common/openapi.yml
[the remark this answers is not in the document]
mmorgantaylor: oh, ha, good find.

service/src/main/java/bio/terra/pipelines/app/controller/JobsApiController.java
```diff
@@ -53,8 +60,8 @@ public ResponseEntity<ApiJobReport> getJob(@PathVariable("jobId") UUID jobId) {
     final SamUser userRequest = getAuthenticatedInfo();
     String userId = userRequest.getSubjectId();
     logger.info("Retrieving jobId {} for userId {}", jobId, userId);
-    FlightState flightState = jobService.retrieveJob(jobId, userId);
-    ApiJobReport result = JobApiUtils.mapFlightStateToApiJobReport(flightState);
+    FlightState flightState = jobService.retrieveJob(jobId, userId, null);
```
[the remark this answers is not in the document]
mmorgantaylor: good call, this was lazy on my part.

service/src/test/java/bio/terra/pipelines/configuration/external/IngressConfigurationTest.java
mmorgantaylor: nice, good call!

service/src/test/java/bio/terra/pipelines/dependencies/stairway/JobServiceMockTest.java
mmorgantaylor: this has been bothering me for a while. also some test names start with "test" and others don't. I'm refactoring them to remove "test" from the names as well as remove snake case.

jsotobroad (approved):
awesome thanks for all the changes

PR #56, opened 2024-02-13
TSPS-170 move all sync calls into async calls using stairway along with tests

jsotobroad:
### Description 

This PR makes it so all of the functionarlity of the service is actually using stairway as opposed to running things in a sync way previously. 

I wasn't 100% sure on what types of exceptions to through when i wanted to fail a step but I took a stab at what seemed reasonable.  The steps may have too much log in them in some places so we should chat about that if we think they should be made more simpler.

Looking at this PR i feel like i want to move the health check to its own steps but not sure if thats over optimizing right now

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-170


service/src/main/java/bio/terra/pipelines/service/ImputationService.java
```diff
@@ -95,11 +70,17 @@ public UUID createImputationJob(
             .newJob()
             .jobId(jobId)
             .flightClass(RunImputationJobFlight.class)
-            .addParameter(JobMapKeys.PIPELINE_NAME.getKeyName(), pipelineName)
+            .addParameter(JobMapKeys.PIPELINE_NAME.getKeyName(), pipelineName.getValue())
```
mmorgantaylor, line 73: why store the String version rather than PipelinesEnum?

service/src/main/resources/application.yml
```diff
@@ -16,15 +16,16 @@ env:
     exportEnabled: ${CLOUD_TRACE_ENABLED:false}
     samplingRate: ${SAMPLING_PROBABILITY:0}
   urls:
-    sam: ${SAM_ADDRESS:https://sam.dsde-dev.broadinstitute.org}
-    leonardo: ${LEONARDO_ADDRESS:https://leonardo.dsde-dev.broadinstitute.org}
+    sam: ${SAM_ADDRESS:https://sam.jsoto-fourth-try.bee.envs-terra.bio}
```
mmorgantaylor, line 19: will you change these back to dev before merging?

service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
```diff
@@ -0,0 +1,88 @@
...
+    recordAttributes.put("scatter", 2);
+    RecordRequest createRecordRequest = new RecordRequest().attributes(recordAttributes);
+    try {
+      wdsService.createOrReplaceRecord(
+          wdsUri,
+          samService.getTspsServiceAccountToken(),
+          createRecordRequest,
+          controlWorkspaceId,
+          pipelineName.getValue(),
+          flightContext.getFlightId(),
```
mmorgantaylor, line 72: it would be helpful to have a comment here noting that we're using the flightId as the primary key in the WDS table

service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
```diff
@@ -0,0 +1,88 @@
...
+import org.databiosphere.workspacedata.model.RecordRequest;
+
+/**
+ * This step creates or replaces a row to a WDS table specific to the pipeline that was launched
+ * currently it writes the flight id as the primary key and a hardcoded "scatter" value that will be
+ * replaced once inputs are being passed in from the user.
+ *
+ * <p>The step expects a cbas uri to be passed in through the working map
+ */
+public class AddWdsRowStep implements Step {
```
mmorgantaylor, line 24: reminder to update the createJob steps sheet with the step name you used and changes to working map updates (e.g. we don't need to store the wds row id because we're using the flight id)

service/src/test/java/bio/terra/pipelines/stairway/PollCromwellRunSetStatusStepTest.java
```diff
@@ -0,0 +1,124 @@
...
+    PollCromwellRunSetStatusStep pollCromwellRunSetStatusStep =
+        new PollCromwellRunSetStatusStep(cbasService, samService, imputationConfiguration);
+    StepResult result = pollCromwellRunSetStatusStep.doStep(flightContext);
+
+    // make sure the step was a success
+    assertEquals(StepStatus.STEP_RESULT_SUCCESS, result.getStepStatus());
+  }
+
+  @Test
+  void doStepNotAllSuccessfulRuns() throws InterruptedException {
```
mmorgantaylor, line 95: great tests!

service/src/test/java/bio/terra/pipelines/stairway/SubmitCromwellRunSetStepTest.java
```diff
@@ -0,0 +1,127 @@
...
+    when(cbasService.checkHealth(any(), any()))
+        .thenReturn(new HealthCheckWorkspaceApps.Result(true, "cbas is healthy"));
+    when(cbasService.getAllMethods(any(), any())).thenReturn(getAllMethodsResponse);
+
+    // do the step
+    SubmitCromwellRunSetStep submitCromwellRunSetStep =
+        new SubmitCromwellRunSetStep(cbasService, samService);
+    StepResult result = submitCromwellRunSetStep.doStep(flightContext);
+
+    // make sure the step was a success
```
mmorgantaylor, line 115: nit: update comment

service/src/test/java/bio/terra/pipelines/stairway/PollCromwellRunSetStatusStepTest.java
```diff
@@ -0,0 +1,124 @@
...
+    when(cbasService.checkHealth(any(), any()))
+        .thenReturn(new HealthCheckWorkspaceApps.Result(true, "cbas is healthy"));
+    when(cbasService.getRunsForRunSet(any(), any(), any())).thenReturn(responseWithErrorRun);
+
+    // do the step
+    PollCromwellRunSetStatusStep pollCromwellRunSetStatusStep =
+        new PollCromwellRunSetStatusStep(cbasService, samService, imputationConfiguration);
+    StepResult result = pollCromwellRunSetStatusStep.doStep(flightContext);
+
+    // make sure the step was a success
```
mmorgantaylor, line 112: nit: update comment

service/src/test/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightTest.java
```diff
@@ -34,24 +30,6 @@ class RunImputationJobFlightTest extends BaseEmbeddedDbTest {
...
-            testJobId,
-            inputParameters,
-            STAIRWAY_FLIGHT_TIMEOUT_SECONDS,
-            /*debugInfo*/ null);
-
-    assertEquals(FlightStatus.SUCCESS, flightState.getFlightStatus());
-  }
-
   @Test
   void createJobFlight_setup() {
```
mmorgantaylor, line 34: here's another snake case test name to fix

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
```diff
@@ -0,0 +1,128 @@
...
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new InternalServerErrorException(
+              "couldn't find method version id for method name " + wdlMethodName));
+    }
+
+    // launch a cbas submission
+    // this is mostly a manually generated run set request definition, we'll want to be able to auto
+    // generate this in the future
+    RunSetRequest runSetRequest =
```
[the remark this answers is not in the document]
jsotobroad: i think im gonna leave this here for now since i want it to be obvious that this is something we want to get rid of in this step.

service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
mmorgantaylor: I personally think it would be helpful - when I was looking at this I was wondering to myself where/how in this call you're specifying the primary key. you're right that I could have figured it out from the java doc but it wouldn't hurt to have that note here. not a huge deal though, I suspect this may get refactored at some point anyway

mmorgantaylor (approved):
super exciting!

service/src/test/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightTest.java
```diff
@@ -53,6 +66,14 @@ void createJobFlightSetup() {
   void expectedStepsInFlight() {
     RunImputationJobFlight runImputationJobFlight =
         new RunImputationJobFlight(StairwayTestUtils.CREATE_JOB_INPUT_PARAMS, flightBeanBag);
-    assertEquals(8, runImputationJobFlight.getSteps().size());
+    assertEquals(expectedStepNames.size(), runImputationJobFlight.getSteps().size());
+
+    Set<String> stepNames =
+        runImputationJobFlight.getSteps().stream()
+            .map(step -> step.getClass().getSimpleName())
```
mmorgantaylor, line 73: awesome

PR #60, opened 2024-03-11
Enable metrics sent to prometheus for api latency metrics

jsotobroad:
### Description 
We want to be able to create a grafana dashboard that tracks API metrics and this should enable that to be possible.
Here is the dashboard - https://grafana.dsp-devops.broadinstitute.org/d/3VhQ1_0Sz/tsps?orgId=1

example of the failed imputation workflow count being recorded in our prometheus endpoint
Screenshot 2024-03-12 at 10 25 15 AM

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-127

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlight.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlight.java
@@ -1,7 +1,9 @@
 package bio.terra.pipelines.stairway.imputation;
 
+import bio.terra.pipelines.app.common.MetricsUtils;
 import bio.terra.pipelines.common.utils.FlightBeanBag;
 import bio.terra.pipelines.common.utils.FlightUtils;
+import bio.terra.pipelines.common.utils.PipelinesEnum;
 import bio.terra.pipelines.dependencies.stairway.JobMapKeys;
 import bio.terra.pipelines.stairway.*;
 import bio.terra.stairway.*;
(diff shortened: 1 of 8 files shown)


service/src/main/resources/application.yml
```diff
@@ -89,7 +89,13 @@ management:
   endpoints:
     web:
       exposure:
-        include: "prometheus"
+        include: "*"
```
mmorgantaylor, line 92: what else is caught with this "*" now?

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlight.java
```diff
@@ -36,6 +38,11 @@ public RunImputationJobFlight(FlightMap inputParameters, Object beanBag) {
         RunImputationJobFlightMapKeys.WDL_METHOD_NAME,
         JobMapKeys.RESULT_PATH.getKeyName());
 
+    PipelinesEnum pipelinesEnum =
+        PipelinesEnum.valueOf(
+            inputParameters.get(JobMapKeys.PIPELINE_NAME.getKeyName(), String.class));
+    MetricsUtils.incrementPipelineRun(pipelinesEnum);
```
mmorgantaylor, line 44: is there a way to test this in `RunImputationJobFlightTest` ?

mmorgantaylor (commented):
couple questions

service/src/main/resources/application.yml
jsotobroad: all of [these](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints) . I have seen other services do this too but yea im not sure we actually need to expose them all.  They arent publicly exposed so its not a security risk or anything.  If anything it would be a performance thing to think about.

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlight.java
jsotobroad: good point, added!

service/src/main/resources/application.yml
mmorgantaylor: would it make sense to add a comment with this link in case we come back in the future and don't remember what this is about? or is it pretty straightforward to find it given the context here, do you think?

mmorgantaylor (approved):
looks awesome!

PR #61, opened 2024-03-14
TSPS-134 Upgrade to Spring Boot 3 and TCL 0.1.9

mmorgantaylor:
### Description 

This PR upgrades TSPS to Spring Boot 3, which involved a few other changes:
- upgrade to terra-common-lib 0.1.9
- upgrade other terra service client libraries to jakarta versions
- note that the new version of Leo's `listAppsV2` endpoint doesn't accept a creatorRoleSpecifier, so we removed it, as we weren't using that functionality anyway
- remove spotbugs (following other services, since we use SonarCloud)
- convert `LeonardoServerConfiguration` from a record to a class, because we were getting an error "Non-canonical record constructor must delegate to another constructor" and converting to a class fixed it
- update the implementation of EmbeddedDataSource for testing based on [this](https://github.com/zonkyio/embedded-database-spring-test/issues/234#issuecomment-1500956434)

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-134


service/build.gradle
```diff
@@ -24,6 +25,14 @@ dependencyManagement {
     }
 }
 
+configurations {
+    // https://stackoverflow.com/questions/76706612/standard-commons-logging-discovery-in-action-with-spring-jcl
+    // resolves "Standard Commons Logging discovery in action with spring-jcl: please remove commons-logging.jar from classpath in order to avoid potential conflicts"
+    configureEach {
+        exclude group: 'commons-logging', module: 'commons-logging'
```
jsotobroad, line 32: would it be worth removing this specifically from each dependency instead of doing it for all our dependencies?

service/build.gradle
```diff
@@ -32,61 +41,75 @@ dependencies {
...
     // Get stairway via TCL
-    implementation "bio.terra:terra-common-lib" // 0.0.94-SNAPSHOT is defined in bio.terra.pipelines.java-common-conventions.gradle
+    implementation "bio.terra:terra-common-lib" // 0.1.9-SNAPSHOT is defined in java-common-conventions
+
+    // terra clients
+    implementation "org.broadinstitute.dsde.workbench:sam-client_2.13:0.1-5a5bf28"
+    implementation "org.broadinstitute.dsde.workbench:leonardo-client_2.13:1.3.6-22ee00b"
+    implementation "org.databiosphere:workspacedataservice-client-okhttp-jakarta:0.2.115-20240227.195850-1"
+    implementation "bio.terra:cbas-client:0.0.199"
     implementation "bio.terra:terra-resource-buffer-client:0.4.3-SNAPSHOT"
```
jsotobroad, line 51: just noticing this is in here, what is this and does it not need to be updated?

service/src/main/java/bio/terra/pipelines/dependencies/sam/SamClient.java
```diff
@@ -30,7 +32,7 @@ private ApiClient getApiClient() {
     var okHttpClientWithTracing =
         this.okHttpClient
             .newBuilder()
-            .addInterceptor(new OkHttpClientTracingInterceptor(Tracing.getTracer()))
+            .addInterceptor(new OkHttpClientTracingInterceptor(openTelemetry))
```
jsotobroad, line 35: can we do this for wds as well? i believe they also use `okHttp` to make their client

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobService.java
```diff
@@ -103,12 +111,12 @@ public void initialize() {
...
             .context(flightBeanBag)
             .addHook(mdcHook)
-            .addHook(new MonitoringHook())
+            .addHook(new MonitoringHook(openTelemetry))
             .exceptionSerializer(new StairwayExceptionSerializer(objectMapper)));
   }
 
   /** Retrieves Job Result specifying the result class type. */
-  @Traced
+  @WithSpan
```
jsotobroad, line 119: are there any docs we can link to like in the readme to explain this tracing stuff?

service/src/main/resources/application.yml
```diff
@@ -61,6 +66,13 @@ spring:
     url: ${env.db.host}/${env.db.pipelines.name}
     username: ${env.db.pipelines.user}
 
+  jpa:
+    # set the following to true to enable SQL logging in JPA
+    show-sql: false
```
jsotobroad, line 71: awesome thanks for the debugging switches!

jsotobroad (commented):
this is great, id love to hear how you fixed the embedded database issues

Also i wonder if we can use the trace stuff in other clients too (to do in another ticket)

.github/workflows/build-and-test.yml
```diff
@@ -46,11 +46,6 @@ jobs:
       - name: Build the test harness and, by dependency, the service library
         run: ./gradlew --build-cache build -x test
 
-      - name: Upload spotbugs results
```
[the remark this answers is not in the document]
mmorgantaylor: yep, a bunch of other services took this out because sonarqube covers these checks. (noted in PR description)

buildSrc/src/main/groovy/bio.terra.pipelines.java-common-conventions.gradle
```diff
@@ -19,12 +18,8 @@ if (!isGithubAction) {
     }
 }
 
-
-java {
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(17)
-    }
-}
+// removing the java toolchain gets rid of a warning about using a deprecated API:
```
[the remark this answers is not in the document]
mmorgantaylor: ah yeah I left this here so we could talk about it more - this is something i'm not sure about

service/build.gradle
mmorgantaylor: probably yes... i was being lazy because when i took it out of the dependency i expected, it didn't get rid of the warning. i'll work on this

service/build.gradle
mmorgantaylor: nice catch, we don't use this - I think I added it by mistake when adding Stairway a while ago. will remove

service/src/main/java/bio/terra/pipelines/dependencies/sam/SamClient.java
mmorgantaylor: definitely we can - are you ok if that goes in the tracing ticket (TSPS-107) rather than this PR?

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobService.java
mmorgantaylor: probably Doug's blog post (which is linked in the tracing ticket), or stuff that's linked in it. do you want me to add that documentation in this PR?

client/swagger.gradle
```diff
@@ -1,12 +1,19 @@
 dependencies {
+    api 'org.springframework.boot:spring-boot-starter-json'
```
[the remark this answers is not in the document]
mmorgantaylor: don't think we do! nice catch

service/build.gradle
mmorgantaylor: offline conversation: leave as-is for now

jsotobroad (approved):
nice this is awesome!!

PR #64, opened 2024-03-15
[TSPS-23] Third party dependency security updates

mmorgantaylor:
### Description 

Per latest sourceclear scan (screenshot in Jira ticket), the following updates:
- logback-classic -> now 1.4.14 (via spring boot -> 3.2.3)
- logback-core -> now 1.4.14 (via spring boot -> 3.2.3)
- postgresql -> now 42.6.1 (direct dependency)
- snakeyaml -> now 2.2 (via terra-common-lib -> 1.0.9)

The following are supposedly brought in by terra-common-lib but don't show up on insightDependency. 
- commons-compress* - [TCL updated this to 1.26.0](https://github.com/DataBiosphere/terra-common-lib/blob/develop/build.gradle#L134)

Unclear whether these changes will also upgrade bouncycastle, since this also doesn't show up on insightDependency.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-23

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -68,7 +68,6 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-validation'
     implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
-    implementation 'org.yaml:snakeyaml:2.0'
 
     // OpenTelemetry utilities for various HTTP clients - copied from WSM
     var openTelemetryVersion = '1.32.0'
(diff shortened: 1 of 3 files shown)


service/build.gradle
```diff
@@ -109,6 +109,10 @@ dependencies {
     testImplementation 'org.awaitility:awaitility'
     testImplementation 'org.testcontainers:postgresql:1.17.3'
 
+    // security-related dependency upgrades of transitive dependencies
+    implementation 'org.yaml:snakeyaml:2.0'
```
jsotobroad, line 113: just OoC, is this section pretty much locking in versions of certain dependencies?  if so would it be possible to put them next to the thing pulling it with a "default" version or is it kind of all over the place?

jsotobroad (approved):
random question but looks good (at least with what little i know about the changes 😆)

service/build.gradle
mmorgantaylor: good point, they're all inherited through tcl - see if you think the reorg makes more sense

PR #67, opened 2024-03-18
TSPS-188 Remove MdcHook in favor of StairwayLoggingHook

mmorgantaylor:
### Description 

Now that we’ve upgraded to TCL 1.x.x, which brings in Stairway 1.0.0, this PR uses use [Stairway 1.0.0's native MDC (mapped diagnostic context) support](https://github.com/DataBiosphere/stairway/pull/125), rather than populating the log MDC itself on flight state transitions. This buys us nicer logs coming out of flights.

This is largely based off of [WSM’s PR to do the same thing](https://github.com/DataBiosphere/terra-workspace-manager/pull/1666), which has a lovely description of what we get.

Here's what our logs look like with this change:
image

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-188


jsotobroad (approved):
Awesome!

PR #70, opened 2024-03-29
TSPS-202 upgrade to TCL version 1.1.4 to resolve k8s breaking change

mmorgantaylor:
### Description 

See https://broadworkbench.atlassian.net/browse/PF-3000 and https://github.com/DataBiosphere/terra-common-lib/pull/167

Tested in my BEE; turned on the IN_KUBERNETES flag and reproduced the "KubePodListener caught exception: java.lang.IllegalArgumentException: The field `volumes` in the JSON string is not defined in the `V1NamespaceSpec` properties" on main; these errors went away (and the service continued to function) when testing on this branch.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-202

### DIFF
--- a/buildSrc/src/main/groovy/bio.terra.pipelines.java-common-conventions.gradle
+++ b/buildSrc/src/main/groovy/bio.terra.pipelines.java-common-conventions.gradle
@@ -25,7 +25,6 @@ java {
 }
 
 repositories {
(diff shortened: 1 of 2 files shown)


jsotobroad (approved):
🥇 thanks for testing this!

PR #72, opened 2024-04-16
TSPS-194 add inputs definitions for imputation

mmorgantaylor:
### Description 

Adds a new db table, pipeline_inputs_definitions, that stores info about each input (the pipeline it belongs to, the name of the input, the type its value should be, and whether it's required). This is joined to the pipeline entity, and these definitions are returned to the user when calling the getPipeline endpoint, like so:

```
{
  "pipelineName": "imputation_beagle",
  "displayName": "Beagle Imputation",
  "description": "Impute missing genotypes using Beagle",
  "type": "imputation",
  "inputs": [
    {
      "name": "multi_sample_vcf",
      "type": "String",
      "isRequired": true
    }
  ]
}
```

Currently nothing new is happening on the createJob submission. Validation of user-provided inputs will be introduced in a different PR. 

Also made the switch from minimac4 to beagle.

Additionally, I refactored the imputation_jobs table to jobs, since it doesn't contain imputation-specific information (and pipeline_id is stored there already).

Deployed and tested in my BEE.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-194


service/src/test/java/bio/terra/pipelines/testutils/TestUtils.java
```diff
@@ -32,6 +31,13 @@ public class TestUtils {
   public static final String TEST_PIPELINE_TYPE_2 = "imputation2";
   public static final String TEST_WDL_URL_2 = "http://nowhere2";
   public static final String TEST_WDL_METHOD_NAME_2 = "methodName2";
+  public static final List<PipelineInputsDefinition> TEST_PIPELINE_INPUTS_DEFINITION_LIST =
+      new ArrayList<>(
+          List.of(
+              new PipelineInputsDefinition(2L, "testRequiredStringInput", "String", true),
```
mmorgantaylor, line 37: update this to 3L

mmorgantaylor: and add test to ensure this pipeline_id matches the requested pipeline

common/openapi.yml
```diff
@@ -502,6 +502,24 @@ components:
...
+      type: string
+      format: string
+
+    PipelineInputType:
+        description: |
+            The type of the input field.
+        type: string
+        format: string
+
+    PipelineInputIsRequired:
```
jsotobroad, line 517: i think this one should go first alphabetically (i kinda hate myself for pointing this out)

service/src/main/resources/db/changesets/20240412-testdata.yaml
```diff
@@ -0,0 +1,18 @@
+databaseChangeLog:
```
jsotobroad, line 1: hmm i wonder if there something we can do with test data stuff so we can keep all of the test seeding changes in one migration file 🤔 .  Im curious what would happen if you put these changes in the other test data file and ran against an already existing database.

service/src/main/resources/db/changesets/20240412-testdata.yaml
```diff
@@ -0,0 +1,18 @@
...
+            columns:
+              - column:
+                  name: job_id
+                  value: deadbeef-dead-beef-deaf-beefdeadbeef
+              - column:
+                  name: user_id
+                  value: testUser
+              - column:
+                  name: pipeline_id
+                  valueComputed: (SELECT id FROM pipelines WHERE name='imputation_beagle')
```
jsotobroad, line 17: oh interesting, might be worth doing something similar in the other test data file to make sure its getting the correct pipeline_id value.  I'm honestly surprised the tests still work after deleting the minimac pipeline from that table

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputsDefinition.java
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.persistence.*;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "pipeline_inputs_definitions")
```
jsotobroad, line 12: nitpick: should this be `pipeline_input_definitions` ?

jsotobroad (commented):
some small nitpicks and some questions around the test seeding data.  This is great

common/openapi.yml
mmorgantaylor: ha! you're totally right - thank you!

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputsDefinition.java
mmorgantaylor: ugh, yes, it should be. will change

service/src/main/resources/db/changesets/20240412-testdata.yaml
mmorgantaylor: agreed it'd be nice to keep them in one file. I'll try.

service/src/main/resources/db/changesets/20240412-testdata.yaml
mmorgantaylor: actually i think we can delete the other test data file entirely

service/src/main/resources/db/changesets/20240412-testdata.yaml
mmorgantaylor: I realized we can always keep just one testdata file, since it's only used in the test context, so live environments and bees and our local env won't ever apply them.

jsotobroad: i did notice that my local environment did have a reference to the original test migration file (which may be wrong because i have accidentally ran tests against my local environment lol). Just wanted to double check that your local environment didnt get upset when changing the name of the original test data file.  It should never apply any of the changes in the file but im not sure if it keeps a reference to it / needs to have that reference exist from then on.

mmorgantaylor: I tried in my BEE and it worked fine so I think it's good. my local environment got messed up when I changed the name of the table from pipeline_inputs_definitions to pipeline_input_definitions so I had to blow that away but otherwise it worked fine

jsotobroad (approved):
all the inputs!

PR #74, opened 2024-04-25
update params step in e2e workflow to get bee-destroy step to always run

jsotobroad:
### Description 
I think we're separating the commands of the init-params step incorrectly.  I think it should be a `|` instead of a `>` https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

### Jira Ticket
No ticket explicitly but related to https://broadworkbench.atlassian.net/browse/TSPS-124


mmorgantaylor (approved):
ah, good catch!!

PR #75, opened 2024-04-25
TSPS-195 Validate user-provided inputs

mmorgantaylor:
### Description 

Add input validation. Currently support five input types: 
- STRING
- INTEGER
- VCF (string ending in .vcf.gz)
- STRING_ARRAY
- VCF_ARRAY

Explanation of current validation strategy is in `IMPLEMENTATION_NOTES.md`.

Made some edits to the GlobalExceptionHandler because some of the handlers weren't feeding through `buildApiErrorReport()`, which is where we are scanning for errors to send to Sentry. Was able to remove some redundant logging as well.

Also updated e2e test to have inputs that will pass this validation: https://github.com/broadinstitute/dsp-reusable-workflows/pull/32

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-195


service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -51,4 +52,87 @@ public Pipeline updatePipelineWorkspaceId(PipelinesEnum pipelineName, UUID works
...
+
+    LinkedHashMap<String, Object> inputsMap = castInputsToMap(inputs);
+
+    ArrayList<String> errorMessages =
+        new ArrayList<>(validateRequiredInputsArePresent(inputDefinitions, inputsMap));
+
+    errorMessages.addAll(validateInputTypes(inputDefinitions, inputsMap));
+
+    if (!errorMessages.isEmpty()) {
+      throw new IllegalArgumentException(String.join("; ", errorMessages));
```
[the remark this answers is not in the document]
mmorgantaylor: couldn't find a nice way to make newlines

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -51,4 +52,87 @@ public Pipeline updatePipelineWorkspaceId(PipelinesEnum pipelineName, UUID works
...
     return pipeline;
   }
+
+  public void validateInputs(PipelinesEnum pipelineName, Object inputs) {
+    Pipeline pipeline = getPipeline(pipelineName);
+    List<PipelineInputDefinition> inputDefinitions = pipeline.getPipelineInputDefinitions();
+
+    // if no inputs are required, nothing to validate
+    if (inputDefinitions.isEmpty()) {
+      return;
```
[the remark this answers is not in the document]
mmorgantaylor: yeah we check elsewhere that all pipelines have defined inputs so i took this out

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -44,13 +44,8 @@ public ResponseEntity<ApiErrorReport> errorReportHandler(ErrorReportException ex
...
   public ResponseEntity<ApiErrorReport> validationExceptionHandler(Exception ex) {
     logger.error("Global exception handler: catch stack", ex);
-    // For security reasons, we generally don't want to include the user's invalid (and potentially
-    // malicious) input in the error response, which also means we don't include the full exception.
-    // Instead, we return a generic error message about input validation.
     String validationErrorMessage =
-        "Request could not be parsed or was invalid: "
-            + ex.getClass().getSimpleName()
-            + ". Ensure that all types are correct and that enums have valid values.";
+        "Request could not be parsed or was invalid: " + ex.getMessage();
```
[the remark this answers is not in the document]
mmorgantaylor: we are now constructing a more useful message to return to the user. we still don't return the value the user provided, but the generic comment is no longer applicable because we are not returning this generic error message.

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,68 @@
...
+
+import bio.terra.common.exception.ValidationException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public enum PipelineInputTypesEnum {
+  STRING("string") {
+    @Override
+    public String cast(String fieldName, Object value) {
+      String stringValue = objectMapper.convertValue(value, String.class);
+      if (stringValue.isBlank()) {
```
mmorgantaylor, line 11: note this does throw an error if `stringValue` is null, so we're counting on nulls having been validated before this

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -51,4 +55,108 @@ public Pipeline updatePipelineWorkspaceId(PipelinesEnum pipelineName, UUID works
...
+   */
+  public List<String> validateInputTypes(
+      List<PipelineInputDefinition> inputDefinitions, Map<String, Object> inputsMap) {
+    ArrayList<String> errorMessages = new ArrayList<>();
+    inputDefinitions.forEach(
+        inputDefinition -> {
+          String inputName = inputDefinition.getName();
+          // we assume we have already generated error messages for null inputs, so we can skip them
+          // here
+          if (inputsMap.containsKey(inputName) && inputsMap.get(inputName) != null) {
```
mmorgantaylor, line 144: alternatively we could just put the null validation in here, but this method was starting to feel heavy

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,112 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
+          }
+        });
+    return errorMessages;
+  }
+
+  /**
+   * Check for extra inputs that are not defined in the pipeline; log a warning if there are
+   * unexpected inputs
+   */
+  public void checkForExtraInputs(
```
mmorgantaylor, line 148: this method isn't tested because testing the logger is a pain but I checked that extra input does get logged at the warn level, like so:
```
2024-04-30T13:32:51.917-04:00 eRG9YDZX  WARN 729 --- [pipelines] [nio-8080-exec-1] b.t.pipelines.service.PipelinesService   : Extra inputs provided: extra_input
```

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
mmorgantaylor: (now we explicitly validate not null right above)

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -51,4 +55,108 @@ public Pipeline updatePipelineWorkspaceId(PipelinesEnum pipelineName, UUID works
...
+
+    LinkedHashMap<String, Object> inputsMap = castInputsToMap(inputs);
+
+    ArrayList<String> errorMessages =
+        new ArrayList<>(validateRequiredInputsArePresent(inputDefinitions, inputsMap));
+
+    errorMessages.addAll(validateNoNulls(inputDefinitions, inputsMap));
+
+    errorMessages.addAll(validateInputTypes(inputDefinitions, inputsMap));
+
```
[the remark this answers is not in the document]
mmorgantaylor: added!

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +189,212 @@ void updatePipelineWorkspaceId() {
...
+    String[] stringArrayWithNull = new String[] {null, "has a null value"};
+
+    return Stream.of(
+        // arguments are: type specification, value to be tested, isRequired, expected errorMessage
+        // basic type checks
+        arguments("Integer", "I am a string", true, commonTypeErrorMessage),
+        arguments("Integer", 2.3, true, commonTypeErrorMessage),
+        arguments("Integer", "2.3", true, commonTypeErrorMessage),
+        arguments("String", "", true, notEmptyErrorMessage),
+        arguments("String", "", false, notEmptyErrorMessage),
```
mmorgantaylor, line 337: this is perhaps a misleading test because we're not running the full validation on these, just the validation of type, and the required status is not taken into account in `validateInputTypes`.

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+      }
+      return stringArray;
+    }
+  };
+  final ObjectMapper objectMapper = new ObjectMapper();
+
+  public abstract <T> T cast(String fieldName, Object value);
+
+  private final String value;
```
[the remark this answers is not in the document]
mmorgantaylor: they weren't; i removed them. good catch

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +189,212 @@ void updatePipelineWorkspaceId() {
...
+    LinkedHashMap<String, Object> inputs = new LinkedHashMap<>();
+    inputs.put("input_name", inputValue);
+
+    assertEquals(
+        errorMessage.formatted(type),
+        pipelinesService.validateInputTypes(inputDefinitions, inputs).get(0));
+  }
+
+  @Test
+  void validateOptionalInputTypeValidAbsent() {
```
jsotobroad, line 391: you could modify this to a parameterized test where you test the different configuration of optional/required and present/absent.

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +189,212 @@ void updatePipelineWorkspaceId() {
...
+  void validateInputTypeNotValid(
+      String type, Object inputValue, Boolean isRequired, String errorMessage) {
+    PipelineInputDefinition inputDefinition =
+        new PipelineInputDefinition(1L, "input_name", type, isRequired);
+    List<PipelineInputDefinition> inputDefinitions = new ArrayList<>(List.of(inputDefinition));
+
+    LinkedHashMap<String, Object> inputs = new LinkedHashMap<>();
+    inputs.put("input_name", inputValue);
+
+    assertEquals(
```
jsotobroad, line 385: another way you could have written this test (or really combining the two above tests) is just asserting whether or not the response from `pipelinesService.validateInputTypes()` was empty or not.  You've already tested the expected error message in the tests above and here you're just making sure what you expect to be valid is valid and what you expect to be invalid is invalid.

so your input arguments would instead be
`String type, Object inputValue, Boolean isRequired, Boolean errorExpected`

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,117 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
+   * Check for extra inputs that are not defined in the pipeline; log a warning if there are
+   * unexpected inputs
+   */
+  public void checkForExtraInputs(
+      Pipeline pipeline,
+      List<PipelineInputDefinition> inputDefinitions,
+      Map<String, Object> inputsMap) {
+    Set<String> expectedInputNames =
+        inputDefinitions.stream().map(PipelineInputDefinition::getName).collect(Collectors.toSet());
+    Set<String> providedInputNames = new HashSet<>(inputsMap.keySet());
```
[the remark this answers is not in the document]
mmorgantaylor: it does, but it links `providedInputNames` to `inputsMap` in such a way that when we then do the set subtraction, `inputsMap` is empty. it seemed safer to ensure that `providedInputNames` is a "new" object, though there may be a better way to do this than what I have

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
mmorgantaylor: oh that's nice, I like it. thanks!

service/src/test/java/bio/terra/pipelines/testutils/TestUtils.java
```diff
@@ -72,8 +72,10 @@ public class TestUtils {
   public static final UUID TEST_NEW_UUID_2 =
       UUID.fromString("deadbeef-dead-beef-bbbb-beefdeadbeef");
 
-  public static final Object TEST_PIPELINE_INPUTS =
-      new LinkedHashMap<>(Map.of("first_key", "first_value"));
+  public static final Map<String, Object> TEST_PIPELINE_INPUTS_MAP =
+      Map.of("first_key", "first_value");
+
+  public static final Object TEST_PIPELINE_INPUTS = (Object) TEST_PIPELINE_INPUTS_MAP;
```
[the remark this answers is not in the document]
mmorgantaylor: ah, this was an artifact of my messing around with a different format for inputs via openapi - I'll revert this. thanks for catching it.

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +189,212 @@ void updatePipelineWorkspaceId() {
...
+    assertTrue(pipelinesService.validateRequiredInputs(inputDefinitions, inputs).isEmpty());
+  }
+
+  private static Stream<Arguments> inputTypeValidationsValid() {
+    return Stream.of(
+        // arguments are: type specification, value to be tested, isRequired
+        arguments("Integer", 123, true),
+        arguments("Integer", "123", true),
+        arguments("String", "I am a string", true),
+        arguments("VCF", "path/to/file.vcf.gz", true),
```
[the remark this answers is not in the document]
mmorgantaylor: yeah, agreed. I can easily add stripping now but I'll wait on the scheme validation bit until we know more about where the inputs are going to be coming from

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +189,212 @@ void updatePipelineWorkspaceId() {
...
+        "Problems with pipelineInputs: multi_sample_vcf is required; new_integer_input must be of type Integer",
+        exception.getMessage());
+  }
+
+  @Test
+  void validateInputsNotAMap() {
+    PipelinesEnum pipelinesEnum = PipelinesEnum.IMPUTATION_BEAGLE;
+    Object inputs = new ArrayList<>(List.of("this", "is", "not", "a", "map"));
+    assertThrows(
+        ValidationException.class, () -> pipelinesService.validateInputs(pipelinesEnum, inputs));
```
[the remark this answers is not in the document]
mmorgantaylor: ah yes, I missed this one, thanks for calling out

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,117 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
+   * Validate that all required inputs are present in the inputsMap
+   *
+   * @param inputDefinitions - list of input definitions for a pipeline
+   * @param inputsMap - map of inputs to validate
+   * @return list of error messages for missing required inputs
+   */
+  public List<String> validateRequiredInputs(
+      List<PipelineInputDefinition> inputDefinitions, Map<String, Object> inputsMap) {
+    ArrayList<String> errorMessages = new ArrayList<>();
+    inputDefinitions.stream()
```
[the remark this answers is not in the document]
mmorgantaylor: definitely, I did think about this, and decided that the cost of looping through a few times didn't outweigh the benefit of keeping the various checks separate. thanks for calling that out though!

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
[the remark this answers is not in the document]
jsotobroad: ahh i see, good point.  Makes sense!

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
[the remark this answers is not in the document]
mmorgantaylor: actually i'm going to wait on stripping as well, since it's not needed for validation but will be needed once we're using the values (i.e. when we create the cbas submission object). I'll add a note in that ticket to implement stripping

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,96 @@
...
+      return stringArray;
+    }
+  };
+  final ObjectMapper objectMapper = new ObjectMapper();
+
+  public abstract <T> T cast(String fieldName, Object value);
+
+  private static void validateNotNullOrEmpty(String fieldName, Object value) {
+    if (value == null) {
+      throw new ValidationException(String.format("%s must not be null", fieldName));
```
nmalfroy, line 84: If this is java 17+, you can do:
```suggestion
      throw new ValidationException("%s must not be null".format(fieldName));
```

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -1,10 +1,15 @@
...
+import bio.terra.common.exception.ValidationException;
+import bio.terra.pipelines.common.utils.PipelineInputTypesEnum;
 import bio.terra.pipelines.common.utils.PipelinesEnum;
 import bio.terra.pipelines.db.entities.Pipeline;
+import bio.terra.pipelines.db.entities.PipelineInputDefinition;
 import bio.terra.pipelines.db.repositories.PipelinesRepository;
-import java.util.List;
-import java.util.UUID;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.*;
```
nmalfroy, line 11: I think generally Terra services avoid `import *` statements.  You can configure IntelliJ to not do that

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,116 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
     pipeline.setWorkspaceId(workspaceId);
     pipelinesRepository.save(pipeline);
     return pipeline;
   }
+
+  public void validateInputs(PipelinesEnum pipelineName, Object inputs) {
+    Pipeline pipeline = getPipeline(pipelineName);
+    List<PipelineInputDefinition> inputDefinitions = pipeline.getPipelineInputDefinitions();
+
+    LinkedHashMap<String, Object> inputsMap = castInputsToMap(inputs);
```
nmalfroy, line 65: Is this a `LinkedHashMap` to preserve order of the keys?  If not, it's generally better to just define it as map 
```suggestion
    Map<> inputsMap = castInputsToMap(inputs);
```

or even 
```suggestion
    var inputsMap = castInputsToMap(inputs);
```

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,116 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
+    ArrayList<String> errorMessages =
+        new ArrayList<>(validateRequiredInputs(inputDefinitions, inputsMap));
+
+    errorMessages.addAll(validateInputTypes(inputDefinitions, inputsMap));
+
+    checkForExtraInputs(pipeline, inputDefinitions, inputsMap);
+
+    if (!errorMessages.isEmpty()) {
+      throw new ValidationException(
+          String.format("Problem(s) with pipelineInputs: %s", String.join("; ", errorMessages)));
```
nmalfroy, line 76: Same thing about format

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,116 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
+    }
+  }
+
+  private LinkedHashMap<String, Object> castInputsToMap(Object inputs) {
+    try {
+      return objectMapper.convertValue(inputs, LinkedHashMap.class);
```
nmalfroy, line 82: Does this need to explicitly return a `LinkedHashMap`?  Could it just return a Map?

Also, to preserve types (in case you get a warning) you do this:
```suggestion
      return objectMapper.convertValue(inputs, new TypeReference<LinkedHashMap<String, Object>(){});
```

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -41,14 +47,116 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
     return pipeline;
   }
+
+  public void validateInputs(PipelinesEnum pipelineName, Object inputs) {
+    Pipeline pipeline = getPipeline(pipelineName);
+    List<PipelineInputDefinition> inputDefinitions = pipeline.getPipelineInputDefinitions();
+
+    LinkedHashMap<String, Object> inputsMap = castInputsToMap(inputs);
+
+    ArrayList<String> errorMessages =
```
nmalfroy, line 67: You could just make this a `List`

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -97,7 +111,7 @@ void addPipelineInput() {
     PipelineInputDefinition newInput = new PipelineInputDefinition();
     newInput.setPipelineId(pipeline.getId());
     newInput.setName("newInput");
-    newInput.setType("Int");
+    newInput.setType("INTEGER");
```
nmalfroy, line 114: Should these be coming from `PipelineInputTypesEnum`?

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +184,171 @@ void updatePipelineWorkspaceId() {
...
+    String notNullErrorMessage = "input_name must not be null";
+    String notEmptyErrorMessage = "input_name must not be empty";
+    String emptyArrayErrorMessage = "input_name must not be an empty array";
+
+    return Stream.of(
+        // arguments: type specification, value to be tested, whether it should pass validation,
+        // expected error message if not
+        arguments("INTEGER", 123, true, null),
+        arguments("INTEGER", "123", true, null),
+        arguments("STRING", "I am a string", true, null),
```
nmalfroy, line 279: Worth verifying that `"\"I am a string\""` returns `"\"I am a string\""` (I'm pretty sure that with the object mapper it'll convert it to `"I am a string"`

nmalfroy: And actually, generally, it might make sense to have some unit testing directly on the  `PipelineInputTypesEnum` class

nmalfroy (commented):
Looks awesome!  I had a couple of comments and I'm happy to go over in person if they don't make sense.  A lot of the stuff might be misunderstanding on my part

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,96 @@
...
+import bio.terra.common.exception.ValidationException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public enum PipelineInputTypesEnum {
+  STRING {
+    @Override
+    public String cast(String fieldName, Object value) {
+      validateNotNullOrEmpty(fieldName, value);
+      String stringValue = objectMapper.convertValue(value, String.class);
+      if (stringValue.isBlank()) {
```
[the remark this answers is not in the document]
mmorgantaylor: hm, I'm confused, I wouldn't expect `" foo  ".isBlank()` to evaluate to true. or am I misunderstanding your concern? 

but in any case, I think we don't need this `if` block anymore - I think it's leftover from when I was doing things slightly differently. ok to cut it out entirely?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,96 @@
+package bio.terra.pipelines.common.utils;
+
+import bio.terra.common.exception.ValidationException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public enum PipelineInputTypesEnum {
+  STRING {
+    @Override
+    public String cast(String fieldName, Object value) {
+      validateNotNullOrEmpty(fieldName, value);
```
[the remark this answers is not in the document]
mmorgantaylor: no longer using objectMapper, but we do test for leading/trailing quotes and they are preserved

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
mmorgantaylor: yes they totally should! nice catch, thanks

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,112 @@
...
+  },
+  STRING_ARRAY {
+    @Override
+    public List<String> cast(String fieldName, Object value) {
+      validateNotNullOrEmpty(fieldName, value);
+      if (value instanceof List<?> listValue) {
+        try {
+          List<String> stringList =
+              listValue.stream()
+                  .map(itemValue -> STRING.cast(fieldName, itemValue).toString())
```
mmorgantaylor, line 58: I had to add the `toString()` to the cast output to make SonarLint happy, but I'm not sure why it's necessary, since STRING.cast() returns a String.

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,110 @@
...
+      } else {
+        throw new ValidationException("%s must be an integer".formatted(fieldName));
+      }
+    }
+  },
+  VCF {
+    @Override
+    public String cast(String fieldName, Object value) {
+      validateNotNullOrEmpty(fieldName, value);
+      if (value instanceof String stringValue) {
```
jsotobroad, line 39: is there a reason we dont use the `STRING` cast function here instead of duplicating it?  Would help keep things consistent if we could

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +195,162 @@ void updatePipelineWorkspaceId() {
...
+  void validateInputType(
+      PipelineInputTypesEnum inputType, Object inputValue, Boolean shouldPassValidation) {
+    PipelineInputDefinition inputDefinition =
+        new PipelineInputDefinition(1L, "input_name", inputType.toString(), true);
+    List<PipelineInputDefinition> inputDefinitions = new ArrayList<>(List.of(inputDefinition));
+
+    LinkedHashMap<String, Object> inputs = new LinkedHashMap<>();
+    inputs.put("input_name", inputValue);
+
+    if (shouldPassValidation) {
```
jsotobroad, line 349: i think you can change this block to 

```
assertEquals(shouldPassValidation, PipelinesService.validateInputTypes(inputDefinitions, inputs).isEmpty())
```

unless you're testing only 1 error gets returned as well

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -170,4 +195,162 @@ void updatePipelineWorkspaceId() {
...
     assertEquals(savedWorkspaceId, p.getWorkspaceId());
   }
+
+  // input validation tests
+  private static Stream<Arguments> inputValidations() {
+    return Stream.of(
+        // arguments: isRequired, inputs, shouldPassValidation, expectedErrorMessage
+        arguments(
+            new LinkedHashMap<String, Object>(
+                Map.of("multi_sample_vcf", "this/is/a/vcf/path.vcf.gz", "new_integer_input", 123)),
```
jsotobroad, line 205: it would probably be a little clearer from a mental gymnastics standpoint and be easier to update if the keys used here were pulled out into variables and used to format the error messages below

jsotobroad (commented):
one refactor suggestion and some nits on tests. looks great though

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
mmorgantaylor: oh nice, thanks!

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
mmorgantaylor: let me know if this update isn't what you had in mind

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
mmorgantaylor: ah yeah I thought about that at some point but didn't do it. great suggestion!

service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+  },
+  VCF {
+    @Override
+    public String cast(String fieldName, Object value) {
+      validateNotNullOrEmpty(fieldName, value);
+      String stringValue = "";
+      try {
+        stringValue = STRING.cast(fieldName, value).toString();
```
jsotobroad, line 41: just noting we had a slack convo about this

jsotobroad (approved):
looks good.  Thanks!!!

PR #78, opened 2024-05-10
update e2e test create-bee template

jsotobroad:
### Description 

We want to use a bee template that has janitor enabled on it to clean up any orphaned resources.  We were asked to create our own template as opposed to updating the rawls one to include tsps.

### Jira Ticket
N/A


mmorgantaylor (approved):
woohoo

PR #79, opened 2024-05-10
Example of how you might do casting

nmalfroy:
Take with a big ol' grain of salt but this seems to do the thing.

Note: I wrapped one of the lists in a `List.of` to make it return the same kind of internal list.  It's really more to make the tests happy.  I think application code would be fine without it.

In any case, let me know what you think!

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
+++ b/service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
@@ -56,11 +56,9 @@ public enum PipelineInputTypesEnum {
       validateNotNullOrEmpty(fieldName, value);
       if (value instanceof List<?> listValue) {
         try {
-          List<String> stringList =
+          List<?> stringList =
               listValue.stream()
-                  .map(
-                      itemValue ->
-                          STRING.cast(fieldName, itemValue, new TypeReference<>() {}).toString())
+                  .map(itemValue -> STRING.cast(fieldName, itemValue, new TypeReference<>() {}))
                   .toList();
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/pipelines/common/utils/PipelineInputTypesEnum.java
```diff
@@ -71,16 +74,25 @@ public List<String> cast(String fieldName, Object value) {
...
+    public <T> T cast(String fieldName, Object value, TypeReference<T> typeReference) {
       validateNotNullOrEmpty(fieldName, value);
       if (value instanceof List<?> listValue) {
         try {
           List<String> stringList =
               listValue.stream()
-                  .map(itemValue -> VCF.cast(fieldName, itemValue).toString())
+                  .map(
+                      itemValue ->
+                          VCF.cast(fieldName, itemValue, new TypeReference<>() {}).toString())
```
mmorgantaylor, line 85: hm you still have `.toString()` here, and that's what we wanted to avoid - when implementing these changes, I still can't remove this

mmorgantaylor: ooh nevermind it works if the typereference input to that cast there is TypeReference<String> rather than TypeReference<>

nmalfroy: Oh whoops, just added a thing to make it behave

mmorgantaylor:
implemented in this PR: https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/81

PR #80, opened 2024-05-13
TSPS-231 update exceptions thrown in flight to be deserializable

jsotobroad:
### Description 

These 3 exceptions are now deserializable when they are thrown in a flight (you can see in the test I changed that the exception we were expecting to fail now succeeds so I changed it to an exception that isnt thrown at all)

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-231

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/stairway/imputation/CheckCbasHealthStep.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/imputation/CheckCbasHealthStep.java
@@ -1,6 +1,5 @@
 package bio.terra.pipelines.stairway.imputation;
 
-import bio.terra.cbas.client.ApiException;
 import bio.terra.pipelines.common.utils.FlightUtils;
 import bio.terra.pipelines.dependencies.cbas.CbasService;
 import bio.terra.pipelines.dependencies.cbas.CbasServiceApiException;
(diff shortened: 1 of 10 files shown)


service/src/main/java/bio/terra/pipelines/stairway/imputation/CheckCbasHealthStep.java
```diff
@@ -36,8 +35,7 @@ public StepResult doStep(FlightContext flightContext)
         cbasService.checkHealth(cbasUri, samService.getTspsServiceAccountToken());
     if (!healthResult.isOk()) {
       return new StepResult(
-          StepStatus.STEP_RESULT_FAILURE_RETRY,
-          new CbasServiceApiException(new ApiException("CBAS is not healthy")));
```
jsotobroad, line 40: it was silly that I committed this in the first place, I shouldn't need to make a whole ApiException object when all i want to pass is a string.  This feels much cleaner

service/src/test/java/bio/terra/pipelines/stairway/imputation/StairwayExceptionSerializerTest.java
```diff
@@ -83,15 +84,16 @@ void deserialize_classNotFound() {
 
   @Test
   void deserialize_failToConstruct() {
-    // this is marked as an ApiErrorReportException, but no matching constructor can be found
+    // this is marked as an ApiErrorReportException, but no matching constructor can be found.  We
+    // dont throw this in our stairway flgiht code so it doesnt need to be deserializable
```
mmorgantaylor, line 88: I think I understand this after thinking about it for a while - to help future devs understand this more easily, can you add some comment here to clarify that we do throw a `LeonardoApiServiceException` but not a `LeonardoServiceException` and why those are different?

mmorgantaylor (approved):
looks great, just one request for comment clarification

can you also show (in the PR description) what the exception looks like now when you run a flight and hit one of these?

service/src/test/java/bio/terra/pipelines/stairway/imputation/StairwayExceptionSerializerTest.java
jsotobroad: updated the comment!

jsotobroad:
here is an example of what the error report for submitting to a workspace the TSPS service acccount doenst have access to looks like 
Screenshot 2024-05-15 at 11 37 52 AM

PR #83, opened 2024-05-16
rename e2e test run name for retrying create bee and attach landing z…

jsotobroad:
### Description 

the current run name where we append `-retry` to the run name doesnt work with the regex the dispatched workflow likes so our retries always fail with this kind of error - https://github.com/broadinstitute/terra-github-workflows/actions/runs/9104110885/job/25027311694.  So just updating the run names so the retries have at least a chance of succeeding

### Jira Ticket
N/A


mmorgantaylor (approved):
nice catch!

PR #84, opened 2024-05-17
TSPS-221 add checker wdl to test if two vcfs are the same

jsotobroad:
### Description 

As part of the work to making beagle determinstic (work done in https://github.com/broadinstitute/warp/pull/1285) we wanted to have wdl that we could pass in two vcfs to and verify that they are in fact the same.

This is the run of this wdl that verifies outputs from two different runs of the new beagle wdl were in fact the same - https://bvdp-saturn-dev.appspot.com/#workspaces/tsps_dev_bp_02_01_2024_v1/imputation-pipeline-testing-03-07-2024/workflows-app/submission-history/14873abd-23c3-4df9-81b8-00d813fd21fe

A run to make sure the workflow does fail when the two files are not the same - https://bvdp-saturn-dev.appspot.com/#workspaces/tsps_dev_bp_02_01_2024_v1/imputation-pipeline-testing-03-07-2024/workflows-app/submission-history/d96627a1-b0eb-4eb9-95ba-841ca19ff2c3

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-221

### DIFF
--- a/pipelines/testing/CompareVcfs.wdl
+++ b/pipelines/testing/CompareVcfs.wdl
@@ -1,6 +1,6 @@
 version 1.0
 
-workflow VerifyVcf {
+workflow CompareVcfs {
 
     input {
         File test_gvcf
--- a/wdls/imputation/VcfChecker.wdl
+++ b/wdls/imputation/VcfChecker.wdl
@@ -25,18 +25,15 @@ task CompareVcfs {
     command {
         set -eo pipefail
 
-        comm --nocheck-order <(gunzip -c -f ~{file1} | grep -v '~{patternForLinesToExcludeFromComparison}')  <(gunzip -c -f ~{file2} | grep -v '~{patternForLinesToExcludeFromComparison}') > comm_output.txt
+        gunzip -c -f ~{file1} | grep -v '~{patternForLinesToExcludeFromComparison}' > file_1.vcf
+        gunzip -c -f ~{file2} | grep -v '~{patternForLinesToExcludeFromComparison}' > file_2.vcf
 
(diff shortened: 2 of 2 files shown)


pipelines/testing/CompareVcfs.wdl
```diff
@@ -0,0 +1,39 @@
...
+        String patternForLinesToExcludeFromComparison
+    }
+    Int disk_size_gb = ceil(3 * size(file1, "GiB")) + ceil(3 * size(file2, "GiB")) + 50
+    command {
+        set -eo pipefail
+
+        gunzip -c -f ~{file1} | grep -v '~{patternForLinesToExcludeFromComparison}' > file_1.vcf
+        gunzip -c -f ~{file2} | grep -v '~{patternForLinesToExcludeFromComparison}' > file_2.vcf
+
+        diff -q file_1.vcf file_2.vcf
```
mmorgantaylor, line 31: is there a way to capture some output to illustrate what the diff is if it fails? it'd be nice to know what the differences are, but currently the log just says 
```Files file_1.vcf and file_2.vcf differ```

jsotobroad: yea we probably wouldnt want to use diff for that cuz the output would be super hard to read.  We'd want another tool to run when we're expecting differences.  This would be mostly meant to make sure things are the same for changes we expect to not make a difference.  (i can remove the `-q` to get the actual differences but it was pretty unreadable to me)

mmorgantaylor: gotcha, makes sense

mmorgantaylor (approved):

PR #86, opened 2024-05-22
TSPS-169 Use user-provided inputs in CBAS run set submission

mmorgantaylor:
### Description 

_Please replace this description with a concise description of this Pull Request._

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-169

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
@@ -7,7 +7,6 @@ import bio.terra.pipelines.common.utils.PipelinesEnum;
 import bio.terra.pipelines.db.entities.Pipeline;
 import bio.terra.pipelines.db.entities.PipelineInputDefinition;
 import bio.terra.pipelines.db.repositories.PipelinesRepository;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -28,7 +27,6 @@ public class PipelinesService {
   private static final Logger logger = LoggerFactory.getLogger(PipelinesService.class);
 
   private final PipelinesRepository pipelinesRepository;
-  ObjectMapper objectMapper = new ObjectMapper();
 
   @Autowired
   public PipelinesService(PipelinesRepository pipelinesRepository) {
@@ -180,13 +178,14 @@ public class PipelinesService {
 
   /**
    * Combine the user-provided inputs map with the service-provided inputs to create a map of all
-   * the inputs for a pipeline.
+   * the inputs for a pipeline. This does not cast the inputs to the correct type, nor does it
+   * format any file inputs with storage container URLs.
    *
    * @param pipelineName - the (enum) name of the pipeline
    * @param userProvidedPipelineInputs - the user-provided inputs
    * @return Map<String, Object> allPipelineInputs - the combined inputs
    */
-  public Map<String, Object> constructInputs(
+  public Map<String, Object> constructRawInputs(
       PipelinesEnum pipelineName, Map<String, Object> userProvidedPipelineInputs) {
 
     Map<String, Object> allPipelineInputs = new HashMap<>(userProvidedPipelineInputs);
--- a/service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
@@ -8,6 +8,8 @@ import bio.terra.pipelines.dependencies.wds.WdsService;
 import bio.terra.pipelines.dependencies.wds.WdsServiceException;
 import bio.terra.stairway.*;
 import bio.terra.stairway.exception.RetryException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.Map;
 import org.databiosphere.workspacedata.model.RecordAttributes;
 import org.databiosphere.workspacedata.model.RecordRequest;
 
@@ -31,6 +33,8 @@ public class AddWdsRowStep implements Step {
   }
 
   @Override
+  @SuppressWarnings("java:S2259") // suppress warning for possible NPE - we do validate not null in
+  // `validateRequiredEntries`
   public StepResult doStep(FlightContext flightContext)
       throws InterruptedException, RetryException {
     // validate and extract parameters from input map
@@ -47,15 +51,23 @@ public class AddWdsRowStep implements Step {
 
     // validate and extract parameters from working map
     FlightMap workingMap = flightContext.getWorkingMap();
-    FlightUtils.validateRequiredEntries(workingMap, RunImputationJobFlightMapKeys.WDS_URI);
+    FlightUtils.validateRequiredEntries(
+        workingMap,
+        RunImputationJobFlightMapKeys.WDS_URI,
+        RunImputationJobFlightMapKeys.ALL_PIPELINE_INPUTS);
 
     String wdsUri = workingMap.get(RunImputationJobFlightMapKeys.WDS_URI, String.class);
+    Map<String, Object> allPipelineInputs =
+        workingMap.get(RunImputationJobFlightMapKeys.ALL_PIPELINE_INPUTS, new TypeReference<>() {});
 
-    // hardcoded for now until we are using inputs from user.
-    // we are using the flight id as the primary key in the table created in WDS
+    // create row to write to WDS
     RecordAttributes recordAttributes = new RecordAttributes();
-    recordAttributes.put("multi_sample_vcf", "a_fake_file.vcf.gz");
-    recordAttributes.put("output_basename", "palantir_42_samples.hg38");
+    recordAttributes.putAll(allPipelineInputs);
+
+    // add a timestamp - TSPS-227 will include generating a real timestamp that we can use here
+    // instead
+    recordAttributes.put("timestamp_start", System.currentTimeMillis());
+
     RecordRequest createRecordRequest = new RecordRequest().attributes(recordAttributes);
     try {
       wdsService.createOrReplaceRecord(
@@ -64,7 +76,7 @@ public class AddWdsRowStep implements Step {
           createRecordRequest,
           controlWorkspaceId,
           pipelineName.getValue(),
-          flightContext.getFlightId(),
+          flightContext.getFlightId(), // this is the primary key for WDS
           "flight_id");
     } catch (WdsServiceException e) {
       // not sure what exception makes sense to throw here
--- a/service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
@@ -9,9 +9,6 @@ import bio.terra.pipelines.dependencies.sam.SamService;
 import bio.terra.pipelines.dependencies.stairway.JobMapKeys;
 import bio.terra.stairway.*;
 import bio.terra.stairway.exception.RetryException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
 import java.util.UUID;
 
 /**
@@ -40,9 +37,11 @@ public class SubmitCromwellRunSetStep implements Step {
     FlightMap inputParameters = flightContext.getInputParameters();
     FlightUtils.validateRequiredEntries(
         inputParameters,
+        JobMapKeys.DESCRIPTION.getKeyName(),
         JobMapKeys.PIPELINE_NAME.getKeyName(),
         RunImputationJobFlightMapKeys.WDL_METHOD_NAME);
 
+    String description = inputParameters.get(JobMapKeys.DESCRIPTION.getKeyName(), String.class);
     PipelinesEnum pipelineName =
         inputParameters.get(JobMapKeys.PIPELINE_NAME.getKeyName(), PipelinesEnum.class);
     String wdlMethodName =
@@ -50,9 +49,15 @@ public class SubmitCromwellRunSetStep implements Step {
 
     // validate and extract parameters from working map
     FlightMap workingMap = flightContext.getWorkingMap();
-    FlightUtils.validateRequiredEntries(workingMap, RunImputationJobFlightMapKeys.CBAS_URI);
+    FlightUtils.validateRequiredEntries(
+        workingMap,
+        RunImputationJobFlightMapKeys.CBAS_URI,
+        RunImputationJobFlightMapKeys.ALL_PIPELINE_INPUTS);
 
     String cbasUri = workingMap.get(RunImputationJobFlightMapKeys.CBAS_URI, String.class);
+    //    Map<String, Object> allPipelineInputs =
+    //        workingMap.get(RunImputationJobFlightMapKeys.ALL_PIPELINE_INPUTS, new
+    // TypeReference<>() {});
 
     // grab methodVersionId needed to submit a submission
     MethodListResponse methodListResponse =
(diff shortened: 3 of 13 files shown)


service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -195,17 +202,62 @@ public Map<String, Object> constructInputs(
...
+                  .inputType(parameterTypeDefinition)
+                  .source(
+                      new ParameterDefinitionRecordLookup()
+                          .recordAttribute(inputName)
+                          .type(ParameterDefinition.TypeEnum.RECORD_LOOKUP));
+            })
+        .toList();
+  }
+
+  protected ParameterTypeDefinition mapInputTypeToCbasParameterType(String type) {
```
mmorgantaylor, line 245: alternatively could add a method in the PipelineInputTypeEnum to return the appropriate ParameterTypeDefinition, but this seems more straightforward

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
```diff
@@ -0,0 +1,87 @@
...
+import bio.terra.stairway.StepResult;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.retry.RetryException;
+
+public class PrepareImputationInputsStep implements Step {
```
mmorgantaylor, line 20: made this imputation-specific

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
```diff
@@ -0,0 +1,87 @@
...
+        inputParameters.get(
+            RunImputationJobFlightMapKeys.USER_PROVIDED_PIPELINE_INPUTS, new TypeReference<>() {});
+
+    Map<String, Object> allPipelineInputs =
+        pipelinesService.constructRawInputs(pipelineEnum, userProvidedPipelineInputs);
+
+    List<PipelineInputDefinition> allInputDefinitions =
+        pipelinesService.getAllPipelineInputDefinitions(pipelineEnum);
+
+    // storage container stuff
```
mmorgantaylor, line 52: there may be a better way to define this

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
```diff
@@ -0,0 +1,87 @@
...
+    List<String> keysToPrependWithStorageURL =
+        List.of("ref_dict", "reference_panel_path", "genetic_maps_path");
+    // in future this will be pulled from the workspace
```
mmorgantaylor, line 55: should make a ticket to do this work (it'll involve adding WSM client)

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -330,4 +336,78 @@ void constructImputationInputsSuccess() {
     }
     assertEquals(totalInputs, allPipelineInputs.size());
   }
+
+  private static Stream<Arguments> mapInputTypeToCbasParameterTypeArguments() {
+    ParameterTypeDefinition stringParameterTypeResponse =
```
mmorgantaylor, line 341: feels like there ought to be a better way to test this, but trying to test the individual components of these objects got real messy real quick

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
```diff
@@ -0,0 +1,87 @@
...
+      PipelineInputTypesEnum pipelineInputType =
+          PipelineInputTypesEnum.valueOf(inputDefinition.getType());
+      String rawValue;
+      if (keysToPrependWithStorageURL.contains(keyName)) {
+        rawValue = workspaceStorageContainerUri + allPipelineInputs.get(keyName).toString();
+      } else {
+        rawValue = allPipelineInputs.get(keyName).toString();
+      }
+      formattedPipelineInputs.put(
+          keyName, pipelineInputType.cast(keyName, rawValue, new TypeReference<>() {}));
```
mmorgantaylor, line 72: if we don't cast here, the real workflow fails in a super ugly way. the empty workflow passes because we don't do anything with the inputs.

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
```diff
@@ -233,11 +144,20 @@ public StepResult doStep(FlightContext flightContext)
...
+            // define the WDS record to link to for inputs and outputs
             .wdsRecords(
                 new WdsRecordSet()
                     .recordType(pipelineName.getValue())
                     .addRecordIdsItem(flightContext.getFlightId()));
 
+    // add inputs
+    List<WorkflowInputDefinition> cbasWorkflowInputDefinitions =
+        pipelinesService.prepareCbasWorkflowInputRecordLookupDefinitions(
+            pipelinesService.getAllPipelineInputDefinitions(pipelineName), wdlMethodName);
```
mmorgantaylor, line 156: would it be better to put this into the inputParameters at the start of the flight, rather than making this call to the TSPS database now?

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
mmorgantaylor: added comment to note this

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -195,17 +202,62 @@ public Map<String, Object> constructInputs(
...
+    return switch (type) {
+      case "STRING", "VCF" -> new ParameterTypeDefinitionPrimitive()
```
jsotobroad, line 247: i wonder if it would be clearer/safer if we converted the String type to a `PipelineInputTypesEnum` for this case statement

service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
```diff
@@ -31,6 +33,8 @@ public AddWdsRowStep(WdsService wdsService, SamService samService) {
   }
 
   @Override
+  @SuppressWarnings("java:S2259") // suppress warning for possible NPE - we do validate not null in
```
jsotobroad, line 36: hmm what was it complaining when using `validateRequiredEntries` that it wasnt complaining about before?

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
jsotobroad: can we add a java doc type explanation of this step

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
jsotobroad: we can either put it in the beginning of the flight or put it in the parameter map when we make the same call in `PrepareImputationInputsStep.java`.  I like the idea of doing it up front since every imputation pipeline will have inputs

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
```diff
@@ -66,112 +72,17 @@ public StepResult doStep(FlightContext flightContext)
...
-                "chr20", "chr21", "chr22"));
-    String refDictInputValue = workspaceStorageContainerUri + "/hg38/Homo_sapiens_assembly38.dict";
-    String geneticMapsPathInputValue =
-        workspaceStorageContainerUri + "/plink-genetic-maps/GRCh38_fixed/";
-    String referencePanelPathInputValue =
-        workspaceStorageContainerUri + "/hg38/Homo_sapiens_assembly38.dict";
-
     // prepare cbas submission (run set) object
-
     // this is mostly a manually generated run set request definition, we'll want to be able to auto
```
jsotobroad, line 76: can remove this comment and the one under it!  Or update it to saw inputs are generated and outputs arent yet generated

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
jsotobroad: i think this is great

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -330,4 +336,78 @@ void constructImputationInputsSuccess() {
...
+  void mapInputTypeToCbasParameterType(String inputType, ParameterTypeDefinition expectedResponse) {
+    assertEquals(expectedResponse, pipelinesService.mapInputTypeToCbasParameterType(inputType));
+  }
+
+  @Test
+  void prepareCbasWorkflowInputRecordLookupDefinitions() {
+    List<PipelineInputDefinition> inputDefinitions = new ArrayList<>();
+    inputDefinitions.add(
+        new PipelineInputDefinition(
+            1L, "input1", PipelineInputTypesEnum.STRING.toString(), true, true, null));
```
jsotobroad, line 377: i would introduce some variability to the `true, true, null` parts of these definitions just to prove they arent actually important (i dont think they are at least)

jsotobroad (commented):
some comments around code structure but looks great!

service/src/main/java/bio/terra/pipelines/stairway/imputation/AddWdsRowStep.java
mmorgantaylor: it's saying `pipelineName` might be null in row 78. but `FlightUtils.validateRequiredEntries` makes sure that each checked entry is not null. of course the checker doesn't know that, though.

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
mmorgantaylor: whoops good catch, thanks!

service/src/main/java/bio/terra/pipelines/stairway/imputation/SubmitCromwellRunSetStep.java
```diff
@@ -233,11 +149,20 @@ public StepResult doStep(FlightContext flightContext)
...
             .wdsRecords(
                 new WdsRecordSet()
                     .recordType(pipelineName.getValue())
                     .addRecordIdsItem(flightContext.getFlightId()));
 
+    // add inputs
+    List<WorkflowInputDefinition> cbasWorkflowInputDefinitions =
+        pipelinesService.prepareCbasWorkflowInputRecordLookupDefinitions(
+            allInputDefinitions, wdlMethodName);
+    for (WorkflowInputDefinition workflowInputDefinition : cbasWorkflowInputDefinitions) {
```
mmorgantaylor, line 162: this `for` block is not covered by tests. think it's worth refactoring all this into a `prepareRunSetRequest` method so we can test this?

service/src/test/java/bio/terra/pipelines/stairway/imputation/AddWdsRowsStepTest.java
```diff
@@ -30,7 +30,13 @@ class AddWdsRowsStepTest extends BaseEmbeddedDbTest {
   void setup() {
     FlightMap inputParameters = new FlightMap();
     FlightMap workingMap = new FlightMap();
+
+    inputParameters.put(
+        RunImputationJobFlightMapKeys.PIPELINE_INPUT_DEFINITIONS, "i'm not the right type");
```
jsotobroad, line 35: is there a reason we're putting an incorrect type here?

jsotobroad (commented):
some small comments and curiousity questions.  this is looking great

service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
```diff
@@ -129,4 +130,14 @@ public boolean equals(Object obj) {
         .append(workspaceId, otherObject.workspaceId)
         .isEquals();
   }
+
+  /**
+   * Get a copy of the pipeline input definitions This is required for this object to be properly
+   * deserialized from the Stairway working map. See second answer here:
+   * https://stackoverflow.com/questions/15833979/java-jackson-deserialize-complex-polymorphic-object-model-jsonmappingexception
+   */
+  @SuppressWarnings("java:S6204") // we do need to stream then collect
```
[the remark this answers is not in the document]
mmorgantaylor: no, it seems to override the Lombok getter (but it doesn't want the @Override annotation). we don't have to change anything about how we call it. basically there's not a way to get the original format. 

another option would be to create a method that returns the "copied" list. do you think that would be less confusing?

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputTypeConverter.java
```diff
@@ -0,0 +1,24 @@
...
+public class PipelineInputTypeConverter
+    implements AttributeConverter<PipelineInputTypesEnum, String> {
+  @Override
+  public String convertToDatabaseColumn(PipelineInputTypesEnum pipelineInputTypeEnum) {
+    return pipelineInputTypeEnum.toString();
+  }
+
+  @Override
+  public PipelineInputTypesEnum convertToEntityAttribute(String pipelineInputTypeString) {
+    return Stream.of(PipelineInputTypesEnum.values())
```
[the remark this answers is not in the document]
mmorgantaylor: I thought about that too, and I didn't think about it too hard but I assumed that this way lets you throw a deliberate exception if your string doesn't match. not sure how useful that is. `return PipelineInputTypesEnum.valueOf(pipelineInputTypeString)` sure seems simpler

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -145,67 +149,103 @@ public List<String> validateInputTypes(
...
+      List<PipelineInputDefinition> inputDefinitions, Map<String, Object> inputsMap) {
     Set<String> expectedInputNames =
         inputDefinitions.stream().map(PipelineInputDefinition::getName).collect(Collectors.toSet());
     Set<String> providedInputNames = new HashSet<>(inputsMap.keySet());
     providedInputNames.removeAll(expectedInputNames);
     if (!providedInputNames.isEmpty()) {
       String concatenatedInputNames = String.join(", ", providedInputNames);
-      logger.warn(
-          "Extra inputs provided for pipeline {}: {}", pipelineName, concatenatedInputNames);
+      logger.warn("Found extra inputs: {}", concatenatedInputNames);
```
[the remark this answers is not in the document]
mmorgantaylor: yeah, to have the pipeline name we'd need to thread it through a couple parent methods. I reasoned that the logs should be identifiable by request so we'll be able to tell which pipeline was being run. but I'm happy to put it back in if you think it's worth it.

service/src/test/java/bio/terra/pipelines/stairway/imputation/AddWdsRowsStepTest.java
mmorgantaylor: whoops, this was leftover from something I was testing, nice catch

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputTypeConverter.java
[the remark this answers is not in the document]
jsotobroad: yea i dont think we'd be creating this entity all that much for our service but its certainly a less efficient way than the `valueOf` route if we were.  Fine to leave as is.

service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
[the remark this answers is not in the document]
jsotobroad: nope if there is only one "getPipelineInputs" function to call thats perfect

jsotobroad (approved):
up to you if you want to test that for loop out, it seems simple enough to not be worried about it.

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputTypeConverter.java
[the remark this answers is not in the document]
mmorgantaylor: easy enough to change to `valueOf`, and so much simpler

PR #87, opened 2024-06-03
TSPS-227 Query pipeline_runs instead of Stairway for pipeline job requests

mmorgantaylor:
### Description 

Rather than using Stairway as the source of truth for TSPS jobs, we want to use our own database. With this PR, when we call createPipelineRun or retrieve a PipelineRunResult, we query the new pipeline_runs table (formerly jobs table) in the TSPS db. 

Corrollary changes:
- write the run to pipeline_runs before kicking off the flight; wrap both in a transaction so if the flight fails to kick off, we don't persist the run in our db either
- add a finalize run step to stairway that sets new "is_success" field to True in the pipeline_runs table
- logic for creating the job result object: if is_success is True, grab everything from the pipeline_runs table, otherwise query Stairway

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-227


service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -183,56 +177,69 @@ public ResponseEntity<ApiCreateJobResponse> createJob(
...
+                  .submitted(pipelineRun.getCreated().toString())
+                  .completed(pipelineRun.getUpdated().toString())
+                  .resultURL(pipelineRun.getResultUrl()))
+          .pipelineOutput(pipelineRun.getOutput());
+    } else {
+      JobApiUtils.AsyncJobResult<String> jobResult =
+          jobService.retrieveAsyncJobResult(
+              pipelineRun.getJobId(), pipelineRun.getUserId(), String.class, null);
+
+      return new ApiCreatePipelineRunResponse()
```
mmorgantaylor, line 237: the following will populate the submitted and completed fields with timestamps from stairway, not from the pipeline_runs table. this means that calling GET pipelines/imputation_beagle/result/{job_id} on a successful run will return different timestamps than calling GET jobs/{job_id}. are we ok with this or should we make them the same? they're off from each other by milliseconds.

service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
```diff
@@ -0,0 +1,262 @@
...
+            testDescription,
+            testPipelineInputs,
+            testResultUrl);
+
+    assertEquals(testJobId, returnedPipelineRun.getJobId());
+    assertEquals(testUserId, returnedPipelineRun.getUserId());
+    assertEquals(testDescription, returnedPipelineRun.getDescription());
+    assertEquals(testResultUrl, returnedPipelineRun.getResultUrl());
+    assertEquals(testPipelineWithId.getId(), returnedPipelineRun.getPipelineId());
+    //    assertNotNull(returnedPipelineRun.getCreated());
```
mmorgantaylor, line 216: these checks fail - I think because we're not out of the transaction at this point so it hasn't actually done the write and generated the timestamps in the db?

common/openapi.yml
```diff
@@ -585,9 +564,9 @@ components:
         type: string
         format: string
 
-    PipelineJobOutput:
+    PipelineRunOutput:
```
jsotobroad, line 567: i think we wanted this section in alphabetical order :D

common/openapi.yml
```diff
@@ -83,8 +83,8 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
     post:
-      summary: Create new pipeline job
-      operationId: createJob
+      summary: Create new pipeline run
+      operationId: createPipelineRun
```
jsotobroad, line 87: do we need to update the integration test script to call this endpoint after it gets merged?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -0,0 +1,187 @@
...
+        new PipelineRun(jobUuid, userId, pipelineId, status.toString(), description, resultUrl);
+    PipelineRun createdPipelineRun = writePipelineRunToDbThrowsDuplicateException(pipelineRun);
+
+    String pipelineInputsAsString;
+    try {
+      // do this to write the pipeline inputs without writing the class name
+      pipelineInputsAsString = objectMapper.writeValueAsString(pipelineInputs);
+    } catch (JsonProcessingException e) {
+      throw new InternalServerErrorException(
+          "THIS SHOULD NEVER HAPPEN! Error converting pipeline inputs to string", e);
```
jsotobroad, line 148: i know this was in there before the chagne but this is probably better left as a log message and not soemthing we might return to the user 😂 .  Probably better to have a differ exception message

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    return pipelineRun;
+  }
+
+  public PipelineRun getPipelineRun(UUID jobId, String userId) {
+    return pipelineRunsRepository.findByJobIdAndUserId(jobId, userId).orElse(null);
+  }
+
+  @Transactional
```
jsotobroad, line 181: why does this need to be transactional?

service/src/main/resources/db/changesets/20240529_convert_jobs_table_to_pipeline_runs.yaml
```diff
@@ -0,0 +1,84 @@
...
+        # note that the liquibase `createTrigger` change type requires a pro license,
+        # so we use the `sql` changeset here
+        - sql:
+            dbms: 'postgresql'
+            sql: >
+              create trigger pipeline_run_updated
+              before update on public.pipeline_runs
+              for each row
+              execute procedure public.pipeline_run_updated_to_now();
+            rollback: drop trigger if exists pipeline_run_updated on public.pipeline_runs;
```
jsotobroad, line 84: \n :)

service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
jsotobroad: it fails here but the `findByJobIdAndUserId` call below returns a pipeline that does ahve a created and updated time? hmmm i wonder if there are annotations we can add to those columns in the PipelineRunEntity so it would make this test work here

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
jsotobroad: So for completedTime if its running it should be null.  If its errored then the user will always see the same time.  If its successful then the user will alway see the same time.  completedTime seems like it'll always be consistent from the users perspective.

For start time the only time it'll change is when the start time they see when its running/failed and the start time they see when its a success.  Its not the best feeling in the world but I'm not sure its worth the effort to bring in the pipeline_runs start time in the `else` part of the code.

common/openapi.yml
mmorgantaylor: discussed on slack - not due to this change - the API path hasn't changed - but we DO want to make a new path, a new group of endpoints for /api/pipelinerun/ rather than having this under /api/pipelines/. thanks!

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: 🤔 not sure why that got in there. it doesn't. good catch, thanks!

common/openapi.yml
mmorgantaylor: this PR description now contains a link to the e2e test change PR

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    return pipelineRun;
+  }
+
+  public PipelineRun getPipelineRun(UUID jobId, String userId) {
+    return pipelineRunsRepository.findByJobIdAndUserId(jobId, userId).orElse(null);
+  }
+
+  public PipelineRun markPipelineRunSuccess(UUID jobId, String userId) {
```
mmorgantaylor, line 180: there's currently no logic here or anywhere to assert that the run actually WAS a success. my thinking was that if we get to the flight step that calls this method, it will only be in the case that the run has succeeded; any error will stop the flight before it gets there. I'm not sure whether to:
1) assume that's the case
2) check here that that's the case and fail if not
3) write the status to SUCCEEDED here regardless of what the previous status was
4) figure out a way to test this

mmorgantaylor: I think that if the logic is, we got to the final step and therefore the flight was a success, then we should be marking the status in our db as SUCCESS here.

mmorgantaylor: discussed offline - added docstring for now

common/openapi.yml
```diff
@@ -114,44 +118,23 @@ paths:
...
-          description: Job is complete (succeeded or failed)
+          description: PipelineRun is complete (succeeded or failed)
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CreateJobResponse'
+                $ref: '#/components/schemas/CreatePipelineRunResponse'
         202:
-          description: Job is running
+          description: PipelineRun is running
```
jsotobroad, line 137: this is funny sounding hahaa

common/openapi.yml
```diff
@@ -114,44 +118,23 @@ paths:
...
+      tags: [ pipelineRuns ]
       responses:
         200:
-          description: Job is complete (succeeded or failed)
+          description: PipelineRun is complete (succeeded or failed)
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CreateJobResponse'
+                $ref: '#/components/schemas/CreatePipelineRunResponse'
```
jsotobroad, line 135: a little off topic but i wonder if we want to rename the `CreatePipelineRunResponse` schema.  It makes sense for the create pipeline api call but not like this result call.  Maybe `AsyncPipelineRunResponse` ? Curious what you think

jsotobroad (approved):
one open question but doesnt block this PR

common/openapi.yml
mmorgantaylor: i wish we could add a 🏃 emoji to this

PR #88, opened 2024-06-04
don't call cache cbas submissions, remove outputs from empty WDL

mmorgantaylor:
### Description 

We don't want to use call caching for now. 
Also, remove outputs that are no longer in the real WDL now that https://github.com/broadinstitute/warp/pull/1288 is merged.

Tested locally:
image


jsotobroad (approved):
awesome thanks for fixing this!

PR #91, opened 2024-06-11
add back in contigs input to beagle empty wdl

jsotobroad:
### Description 

i deleted an input in the empty beagle wdl so this is adding it back in!!


mmorgantaylor (approved):
thank you!!

PR #94, opened 2024-06-20
TSPS-246 prepare pipeline run API endpoint

mmorgantaylor:
### Description 

Caller provides: job id (uuid), pipeline version, and pipeline inputs
Teaspoons creates write sas tokens for all file inputs, writes the pipeline run to the teaspoons db, and returns the (same) jobId + sas urls to user

example payload:
```
{
  "jobId": "22222237-5717-4562-b3fc-2c963f66afa6",
  "pipelineVersion": "string",
  "pipelineInputs": {
    "outputBasename": "foobar",
    "multiSampleVcf": "some/local/path/file.vcf.gz"
  }
}
```
response:
```
{
  "jobId": "22222237-5717-4562-b3fc-2c963f66afa6",
  "pipelineFileInputs": {
    "multiSampleVcf": {
      "sasUrl": "https://lze96253b07f13c61ef712bb.blob.core.windows.net/sc-68a43bd8-e744-4f1e-87a5-c44ecef157a3/user-input-files/22222237-5717-4562-b3fc-2c963f66afa6/file.vcf.gz?sv=2023-11-03&spr=https&st=2024-06-21T20%3A20%3A08Z&se=2024-06-22T20%3A35%3A08Z&sr=b&sp=w&sig=[REDACTED]&rscd=2698170027631a5a65e71",
      "azcopyCommand": "azcopy copy some/local/path/file.vcf.gz https://lze96253b07f13c61ef712bb.blob.core.windows.net/sc-68a43bd8-e744-4f1e-87a5-c44ecef157a3/user-input-files/22222237-5717-4562-b3fc-2c963f66afa6/file.vcf.gz?sv=2023-11-03&spr=https&st=2024-06-21T20%3A20%3A08Z&se=2024-06-22T20%3A35%3A08Z&sr=b&sp=w&sig=[REDACTED]%3D&rscd=2698170027631a5a65e71"
    }
  }
}
```

And confirmed that that SAS url can be used to write a file to that location; cannot read that file back (this is the intended behavior).

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-246


service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -60,6 +65,78 @@ public PipelineRunsService(
...
+    if (Boolean.TRUE.equals(pipelineRunsRepository.existsByJobId(jobId))) {
+      throw new DuplicateObjectException(
+          "Duplicate jobId %s found. Please resubmit with a different jobId".formatted(jobId));
+    }
+
+    // define the path for each file input as the jobId + the input key name
+    List<String> userProvidedInputFileKeys =
+        pipeline.getPipelineInputDefinitions().stream()
+            .filter(PipelineInputDefinition::getUserProvided)
+            .filter(p -> p.getType().equals(PipelineInputTypesEnum.VCF))
```
mmorgantaylor, line 100: at this point we may want to abandon the "VCF" specific type and just use "FILE" since we're not going to be validating suffixes anyway

common/openapi.yml
```diff
@@ -642,6 +669,30 @@ components:
         inputs:
           $ref: "#/components/schemas/PipelineUserProvidedInputDefinitions"
 
+    PreparePipelineRunRequestBody:
+      description: |
+        Object containing the user-provided information defining a pipeline run request.
+      type: object
+      required: [ pipelineVersion, pipelineInputs ]
```
jsotobroad, line 676: we probably need jobId required here

common/openapi.yml
```diff
@@ -83,6 +83,33 @@ paths:
...
+      tags: [ pipelineRuns ]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PreparePipelineRunRequestBody'
+      responses:
+        200:
+          description: File upload SAS URLs
```
jsotobroad, line 101: where do these descriptions end up?  a lot of our 200s just have `Success`  so im curious lol

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
jsotobroad: i think we want to filter on `VCF` and `File` either way since we'd want to generate sas tokens for both.  Is there really not a futre where we'll have the ability to validate a users input file names at the level of the code?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -60,6 +64,84 @@ public PipelineRunsService(
...
+            .collect(
+                HashMap::new,
+                (m, p) ->
+                    m.put(
+                        p.getName(),
+                        "pipeline-run-user-input-files/%s/%s"
+                            .formatted(jobId, p.getDefaultValue())),
```
jsotobroad, line 106: this is definitely weird to have a user provided input where we just use the default value.  maybe we should make the user pass in a file name so that way we can treat it like ther user input that its meant to be?

mmorgantaylor: it's only the default value when it comes to naming. i wouldn't know how to enforce that name when they upload. e.g. what if they specify the input "my_first_file.vcf.gz" at prepareRun, but then upload "my_second_file.txt" to that path?

common/openapi.yml
mmorgantaylor: they show up in the example responses in Swagger. like this: 
image

agree that this particular description isn't quite right 😂

service/src/main/java/bio/terra/pipelines/stairway/imputation/PrepareImputationInputsStep.java
```diff
@@ -61,7 +61,7 @@ public StepResult doStep(FlightContext flightContext) {
 
     // define input file paths that need to be prepended with the workspace storage URL
     List<String> keysToPrependWithStorageURL =
-        List.of("ref_dict", "reference_panel_path_prefix", "genetic_maps_path");
+        List.of("refDict", "referencePanelPathPrefix", "geneticMapsPath");
```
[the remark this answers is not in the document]
mmorgantaylor: not exactly the same, but I added them to the application.yml, wdyt?

service/src/main/java/bio/terra/pipelines/app/common/MetricsUtils.java
```diff
@@ -16,6 +16,20 @@ public class MetricsUtils {
...
 
+  /**
+   * increments metrics counter for a tsps pipeline that has been prepared
+   *
+   * @param pipelineName - name of pipeline that was prepared e.g. "imputation"
+   */
+  public static void incrementPipelinePrepareRun(PipelinesEnum pipelineName) {
+    Metrics.globalRegistry
+        .counter(
+            String.format("%s.pipeline.prepare.count", NAMESPACE),
```
jsotobroad, line 27: can we update this to `pipeline.prepareRun.count`

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -91,6 +91,7 @@ public void validateUserProvidedInputs(
     checkForExtraInputs(userProvidedInputDefinitions, inputsMap);
 
     if (!errorMessages.isEmpty()) {
+      errorMessages.sort(String::compareToIgnoreCase); // sort to help with testing
```
jsotobroad, line 94: hmm we shouldnt be doing this if its just for testing.  In the test we should just assert what we can.

service/src/test/java/bio/terra/pipelines/configuration/internal/ImputationConfigurationTest.java
```diff
@@ -0,0 +1,24 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import bio.terra.pipelines.app.configuration.internal.ImputationConfiguration;
+import bio.terra.pipelines.testutils.BaseEmbeddedDbTest;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+class ImputationConfigurationTest extends BaseEmbeddedDbTest {
```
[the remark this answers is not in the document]
mmorgantaylor: yeah it's because we're using @Autowired ... all our configuration tests use `BaseEmbeddedDbTest`. we could create a `BaseSpringBootTest` but we'd still need to configure a db setup in that case, so not sure it's worth it

PR #95, opened 2024-06-25
TSPS-247 update CreatePipelineRun -> StartPipelineRun

mmorgantaylor:
### Description 

Now that we have a PreparePipelineRun endpoint, we need to wire it through so the user can kick off the pipelineRun that they have prepared (after uploading their input files via azcopy). 

Changes:
- rename createPipelineRun resources to startPipelineRun
- no longer accept pipelineInputs or pipelineVersion when calling createPipelineRun
- add workspaceStorageUrl field to PipelineRun entity (db change)
- in preparePipelineRun, extract the control workspace storage url and store it in the db
- in startPipelineRun, make sure the provided jobId already exists for that user and that its status is PREPARING
- in startPipelineRun, add workspaceStorageUrl to the flight's input parameters
- in PrepareImputationInputsStep, pull workspaceStorageUrl from input params to construct full paths for user-provided input files

Accompanying update to e2e test: https://github.com/broadinstitute/dsp-reusable-workflows/pull/42

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-247


service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -148,78 +157,58 @@ private Map<String, Map<String, String>> prepareFileInputs(
...
+              sasUrl,
+              "azcopyCommand",
+              "azcopy copy %s %s".formatted(fileInputValue, sasUrl)));
     }
 
     return fileInputsMap;
   }
 
-  private Map<String, String> constructSasUrlAndAzcopyCommand(
+  private String retrieveWriteOnlySasUrl(
```
mmorgantaylor, line 177: should this method live in WorkspaceManagerService instead?

service/src/main/java/bio/terra/pipelines/common/utils/FileUtils.java
```diff
@@ -25,6 +31,41 @@ public static String getBlobNameFromTerraWorkspaceStorageHttpUrl(
...
+   * Construct the destination blob name for a user-provided file input.
+   *
+   * <p>For example, file `local/path/to/file.txt` for jobId `1234` returns
+   * `user-input-files/1234/file.txt`
+   *
+   * @param jobId
+   * @param userProvidedFileInputValue
+   * @return blobName
+   */
+  public static String constructBlobNameForUserInputFile(
```
jsotobroad, line 44: would rename to `constructDestinationBlobNameForUserInputFile` based on your description of it.

service/src/main/java/bio/terra/pipelines/common/utils/FileUtils.java
```diff
@@ -25,6 +31,41 @@ public static String getBlobNameFromTerraWorkspaceStorageHttpUrl(
...
+   * Extract the base storage url from a full path or SAS url, using the workspaceId as a delimiter.
+   *
+   * <p>For example, `https://lz123.blob.core.windows.net/sc-{workspaceId}/path/to/file` becomes
+   * `https://lz123.blob.core.windows.net/sc-{workspaceId}`
+   *
+   * @param sasUrl
+   * @param workspaceId
+   * @return baseStorageUrl
+   */
+  public static String getBaseStorageUrlFromSasUrl(String sasUrl, UUID workspaceId) {
```
jsotobroad, line 60: maybe `getStorageContainerUrlFromSasUrl` ?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -124,18 +135,16 @@ public Map<String, Map<String, String>> preparePipelineRun(
    * <p>This SAS url along with the source file path provided by the user are used to generate an
    * azcopy command that the user can run to upload the file to the location in the pipeline
    * workspace storage container.
+   *
+   * <p>The TSPS database will auto-generate created and updated timestamps, so we do not need to
```
jsotobroad, line 139: not sure this comment needs to exist in general but i dont think it has anything to do with the function its commenting on as well

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
jsotobroad: yea i would probably move it there or WSM utils next to the WorkspaceManagerService similar to the AppUtils file for cbas

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -248,26 +237,38 @@ public PipelineRun createPipelineRun(
...
     jobBuilder.submit();
 
-    logger.info("Created {} pipelineRun with jobId {}", pipelineName, jobId);
+    logger.info("Started {} pipelineRun with jobId {}", pipelineName, jobId);
 
     return pipelineRun;
   }
 
+  private Map<String, Object> retrievePipelineInputs(Long id) {
+    PipelineInput pipelineInput = pipelineInputsRepository.findById(id).orElseThrow();
```
jsotobroad, line 255: should we put something specific to throw here if the inputs dont exist in the database?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -93,21 +96,29 @@ public Map<String, Map<String, String>> preparePipelineRun(
       throw new InternalServerErrorException("%s workspaceId not defined".formatted(pipelineName));
     }
 
+    if (getPipelineRun(jobId, userId) != null) {
+      throw new BadRequestException("JobId %s already exists".formatted(jobId));
```
jsotobroad, line 100: should we put something here like "use the x endpoint to get details for it" or something?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -148,78 +157,58 @@ private Map<String, Map<String, String>> prepareFileInputs(
 
     Map<String, Map<String, String>> fileInputsMap = new HashMap<>();
     for (String fileInputName : fileInputNames) {
+      String fileInputValue = userProvidedInputs.get(fileInputName).toString();
```
jsotobroad, line 160: i think its safer to do `(String) userProvidedInputs.get(fileInputName)` if you're looking to cast a value

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -93,21 +96,29 @@ public Map<String, Map<String, String>> preparePipelineRun(
       throw new InternalServerErrorException("%s workspaceId not defined".formatted(pipelineName));
```
jsotobroad, line 96: i was looking at this and we should probably change the `name` column of Pipelines to be auto converted to a PipelinesEnum similar to how we treat the `type` column in the PipelineInputDefinitions table.  We do a lot of conversions back and forth throughout this file and it would just be clearner.  Maybe worth make a ticket for?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -148,78 +157,58 @@ private Map<String, Map<String, String>> prepareFileInputs(
...
+   * Start a PipelineRun that exists in the database (via preparePipelineRun).
    *
    * <p>We encase the logic here in a transaction so that if the submission to Stairway fails, we do
-   * not persist the entry in our own pipeline_runs table.
+   * not update the status in our own pipeline_runs table.
    *
-   * <p>The TSPS database will auto-generate created and updated timestamps, so we do not need to
-   * specify them when writing to the database. The generated timestamps will be included in the
-   * returned PipelineRun object.
+   * <p>The TSPS database will auto-generate updated timestamps.
```
jsotobroad, line 194: should we start using Teaspoons now?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: great idea. https://broadworkbench.atlassian.net/browse/TSPS-263

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: love it

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: nice catch, copy pasta mistake

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: sure - this really shouldn't happen - so we can throw an Internal Server Error with a somewhat descriptive message

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -93,21 +94,31 @@ public Map<String, Map<String, String>> preparePipelineRun(
       throw new InternalServerErrorException("%s workspaceId not defined".formatted(pipelineName));
     }
 
+    if (getPipelineRun(jobId, userId) != null) {
```
mmorgantaylor, line 97: note that this only throws if a jobId already exists that this user submitted. if another user had already submitted a job with this jobId, this currently won't trigger an exception until we attempt to write the job to the db, at which point it will throw a DuplicateObjectException (still a 400 Bad Request). 

should we instead check for any jobId matching this one? in that case the calling user wouldn't be able to retrieve details for it. 

the existing behavior is fine, but it is inefficient in the case that the jobId does exist from another user's request, in that we'll call WSM to retrieve SAS urls before running into the error at db write time.

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -317,10 +329,47 @@ protected PipelineRun writePipelineRunToDbThrowsDuplicateException(PipelineRun p
...
+      throw new BadRequestException(
+          "JobId %s not found. You must prepare a pipeline run before starting it."
+              .formatted(jobId));
+    }
+    // only allow starting a pipeline run if it is in the PREPARING state
+    if (!pipelineRun.getStatus().equals(CommonPipelineRunStatusEnum.PREPARING.toString())) {
+      throw new BadRequestException(
+          "JobId %s is not in the PREPARING state. Cannot start pipeline run.".formatted(jobId));
+    }
+    pipelineRun.setStatus(CommonPipelineRunStatusEnum.RUNNING.toString());
```
jsotobroad, line 366: this column should also get the `Enum` Treatment to make this less annoying. If you wnat to add it to the ticket you created for that other column we can punt on this.

service/src/main/resources/db/changesets/20240625_add_workspace_storage_url_to_pipeline_runs.yaml
```diff
@@ -0,0 +1,12 @@
...
+databaseChangeLog:
+  - changeSet:
+      id: Add a workspace_storage_url column to the pipeline_runs table
+      author: mma
+      changes:
+        - addColumn:
+            tableName: pipeline_runs
+            columns:
+              - column:
+                  name: workspace_storage_url
```
jsotobroad, line 11: extreeeeeeeeeme nitpick so feel free to ignore, `workspace_storage_container_url`

jsotobroad (approved):
couple of nitpicks that can either be put into another ticket or ignored outright, woooo full end to end service working!!

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: good call, updated https://broadworkbench.atlassian.net/browse/TSPS-263

PR #96, opened 2024-06-25
TSPS-174 rename tsps to teaspoons for all cosmetic changes in the repo

jsotobroad:
### Description 

We wanted to break up the renaming changes for the service into cosmetic and non cosmetic changes.  This is the set of changes for the cosmetic stuff

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-174


common/openapi.yml
```diff
@@ -453,15 +453,15 @@ components:
     JobResult:
       type: object
       description: |
-        The result of an async call that triggers a stairway job in TSPS.
+        The result of an async call that triggers a stairway job in Teaspoons.
```
mmorgantaylor, line 456: since this is user-facing text, we probably don't want to mention Teaspoons, and I don't think we really need to. how about we change this line to something like "The result of a call that triggers an asynchronous job"

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelinesSpringConfiguration.java
```diff
@@ -20,8 +20,8 @@
 @Configuration
 public class PipelinesSpringConfiguration {
   @Bean
-  @ConfigurationProperties("tsps.version")
-  public ApiVersionProperties getTspsVersion() {
+  @ConfigurationProperties("teaspoons.version")
+  public ApiVersionProperties getTeaspoonsVersion() {
```
mmorgantaylor, line 24: for some reason this particular one made me happy

service/src/main/java/bio/terra/pipelines/dependencies/sam/SamService.java
```diff
@@ -65,15 +65,15 @@ public ApiSystemStatusSystems checkHealthApiSystemStatus() {
...
+  public String getTeaspoonsServiceAccountToken() {
     try {
       GoogleCredentials creds =
           GoogleCredentials.getApplicationDefault().createScoped(SAM_OAUTH_SCOPES);
       creds.refreshIfExpired();
       return creds.getAccessToken().getTokenValue();
     } catch (IOException e) {
       throw new InternalServerErrorException(
-          "Internal server error retrieving TSPS credentials", e);
+          "Internal server error retrieving Teaspoons credentials", e);
```
mmorgantaylor, line 76: since this could get returned to the user, what if we just say "Internal server error retrieving service credentials"

mmorgantaylor (approved):
looks awesome! just a couple suggestions for changing user-facing text

common/openapi.yml
jsotobroad: love it

service/src/main/java/bio/terra/pipelines/dependencies/sam/SamService.java
jsotobroad: 👍

PR #98, opened 2024-07-09
TSPS-263 auto enum conversion for PipelineRun status and Pipeline name

mmorgantaylor:
### Description 

Rather than doing a bunch of conversions between strings and enums in our code, better to auto-convert directly in/out of the database.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-263


service/src/main/java/bio/terra/pipelines/db/entities/PipelineNameConverter.java
```diff
@@ -0,0 +1,19 @@
+package bio.terra.pipelines.db.entities;
+
+import bio.terra.pipelines.common.utils.PipelinesEnum;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+// inspired by https://www.baeldung.com/jpa-persisting-enums-in-jpa
+@Converter(autoApply = true)
+public class PipelineNameConverter implements AttributeConverter<PipelinesEnum, String> {
```
jsotobroad, line 9: now that we're getting a number of these converter classs, can we put them in their own directory under `/db/` or `/db/entities/`

jsotobroad (approved):
one small nitpick but i think this looks cleaner overall

PR #101, opened 2024-07-10
TSPS-267 actually use our status service

mmorgantaylor:
### Description 

We weren't actually using our StatusService, now we are. This includes checking that Sam is healthy as well as our database connection.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-267

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/service/StatusServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/StatusServiceTest.java
@@ -44,6 +44,13 @@ class StatusServiceTest extends BaseTest {
   }
 
   @Test
+  void testStatusConfigurationNotEnabled() {
+    when(configuration.enabled()).thenReturn(false);
+    statusService.checkStatus();
+    assertTrue(statusService.getCurrentStatus());
+  }
+
+  @Test
   void testStatusOk() {
     doReturn(true).when(jdbcTemplateMock).execute(any(ConnectionCallback.class));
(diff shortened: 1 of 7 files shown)


jsotobroad (approved):

service/src/test/java/bio/terra/pipelines/service/StatusServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+  @MockBean private StatusCheckConfiguration configuration;
+  @MockBean private SamService samService;
+
+  @Mock private final JdbcTemplate jdbcTemplateMock = new JdbcTemplate();
+
+  @BeforeEach
+  void setup() {
+    // by default, everything should work. we can override some of these in specific tests
+    when(jdbcTemplate.getJdbcTemplate()).thenReturn(jdbcTemplateMock);
+    when(configuration.enabled()).thenReturn(true);
```
jsotobroad, line 41: i assume you mocked this out instead of pulling it from the test application.yml so you could make this a `BaseTest` or was there another reason?

jsotobroad (approved):

service/src/test/java/bio/terra/pipelines/service/StatusServiceTest.java
mmorgantaylor: it was easier to change the values via mocking since the configuration doesn't have setters. the speed of using `BaseTest` over `BaseEmbeddedDbTest` is also great but wasn't my motivation originally

PR #102, opened 2024-07-11
TSPS-197 define outputs in db

mmorgantaylor:
### Description 

This PR removes all hardcoding of pipeline output information, instead storing the output definitions in our database. We also move the actual pipeline outputs from the `pipeline_runs` table to a dedicated `pipeline_outputs` table.

Accompanying PR to update the e2e test: https://github.com/broadinstitute/dsp-reusable-workflows/pull/43

Follow-up ticket to harmonize input and output definition types: TSPS-268

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-197


common/openapi.yml
```diff
@@ -486,8 +486,8 @@ components:
           $ref: '#/components/schemas/JobReport'
         errorReport:
           $ref: '#/components/schemas/ErrorReport'
-        pipelineOutput:
-          $ref: '#/components/schemas/PipelineRunOutput'
+        pipelineOutputs:
```
mmorgantaylor, line 489: reminder to self that this will require an update of the integration test

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputs.java
```diff
@@ -13,7 +13,7 @@
 @Setter
 @NoArgsConstructor
 @Table(name = "pipeline_inputs")
-public class PipelineInput {
+public class PipelineInputs {
```
jsotobroad, line 16: oh i think the interface with the API should be PipelineInputs but I think this should still be `PipelineInput`.  Same for the `PipelineOutput` entity

service/src/main/java/bio/terra/pipelines/db/entities/PipelineOutputDefinition.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "pipeline_output_definitions")
+public class PipelineOutputDefinition extends BasePipelineVariableDefinition {
+
+  @Column(name = "type", nullable = false)
+  private String type;
```
jsotobroad, line 18: i think this should be an enum thats share with `PipelineInputDefinition`.  so probalby rename `PipelineInputTypesEnum` to `PipelineypesEnum` and use them in both?  what do you think

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -259,4 +263,31 @@ protected ParameterTypeDefinition mapInputTypeToCbasParameterType(PipelineInputT
...
+   */
+  public List<WorkflowOutputDefinition> prepareCbasWorkflowOutputRecordUpdateDefinitions(
+      List<PipelineOutputDefinition> pipelineOutputDefinitions, String wdlMethodName) {
+    return pipelineOutputDefinitions.stream()
+        .map(
+            pipelineOutputDefinition -> {
+              String outputName = pipelineOutputDefinition.getWdlVariableName();
+              return new WorkflowOutputDefinition()
+                  .outputName("%s.%s".formatted(wdlMethodName, outputName))
+                  .outputType(
```
jsotobroad, line 282: shouldnt this be based on what the type the pipelineOutputDefinition is?

service/src/main/java/bio/terra/pipelines/db/entities/PipelineInputs.java
mmorgantaylor: hm, even though each instance of this entity contains the same information as the API interface, i.e. multiple outputs?

service/src/main/java/bio/terra/pipelines/db/entities/PipelineOutputDefinition.java
mmorgantaylor: probably yes eventually, but for now since there's only one output type, I was keeping it as FILE and leaving it at that, because I didn't want to add FILE to `PipelineInputTypesEnum` and then have the FILE vs VCF discrepancy to leave or resolve in this PR. I'd prefer to make a new ticket to change `VCF` to `FILE` in `PipelineInputTypesEnum` (which would involve adding a field for the suffix to check for input validation) (and rename `PipelineInputTypesEnum` to `PipelineVariableTypesEnum`). thoughts?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: yes, but currently there's only one type (FILE) so until we added others I was keeping it simple. when we switch to using the type enum for outputs, then we'll want to add that conditional logic

service/src/main/java/bio/terra/pipelines/db/entities/PipelineOutputDefinition.java
mmorgantaylor: https://broadworkbench.atlassian.net/browse/TSPS-268

jsotobroad (approved):

PR #104, opened 2024-07-16
TSPS-268 unify PipelineVariableTypesEnum and associated changes

mmorgantaylor:
### Description 

- rename `PipelineInputTypesEnum` to `PipelineVariableTypesEnum`
- use it in base `BasePipelineVariableDefinition` class, so used for both `PipelineInputDefinition` and `PipelineOutputDefinition`
- add a `file_suffix` field to pipeline_input_definitions table in the db, change existing `VCF` types to `FILE`s and add `".vcf.gz"` as the file suffix for those
- refactor enum values `VCF` -> `FILE` and `VCF_ARRAY` -> `FILE_ARRAY`; update the `validate()` method to check the fileSuffix for FILE and FILE_ARRAY types

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-268

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
+++ b/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
@@ -1,5 +1,6 @@
 package bio.terra.pipelines.common.utils;
 
+import bio.terra.pipelines.db.entities.PipelineInputDefinition;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -20,7 +21,8 @@ public enum PipelineVariableTypesEnum {
     }
 
     @Override
-    public String validate(String fieldName, String fileSuffix, Object value) {
+    public String validate(PipelineInputDefinition pipelineInputDefinition, Object value) {
+      String fieldName = pipelineInputDefinition.getName();
       if (cast(fieldName, value, new TypeReference<String>() {}) == null) {
         return "%s must be a string".formatted(fieldName);
       }
@@ -43,7 +45,8 @@ public enum PipelineVariableTypesEnum {
     }
 
     @Override
-    public String validate(String fieldName, String fileSuffix, Object value) {
+    public String validate(PipelineInputDefinition pipelineInputDefinition, Object value) {
+      String fieldName = pipelineInputDefinition.getName();
       if (cast(fieldName, value, new TypeReference<Integer>() {}) == null) {
         return "%s must be an integer".formatted(fieldName);
       }
@@ -57,7 +60,9 @@ public enum PipelineVariableTypesEnum {
     }
 
     @Override
-    public String validate(String fieldName, String fileSuffix, Object value) {
+    public String validate(PipelineInputDefinition pipelineInputDefinition, Object value) {
+      String fieldName = pipelineInputDefinition.getName();
+      String fileSuffix = pipelineInputDefinition.getFileSuffix();
       String stringCastValue = STRING.cast(fieldName, value, new TypeReference<>() {});
       if (stringCastValue == null || !(stringCastValue.endsWith(fileSuffix))) {
         return "%s must be a path to a file ending in %s".formatted(fieldName, fileSuffix);
@@ -85,7 +90,8 @@ public enum PipelineVariableTypesEnum {
     }
 
     @Override
-    public String validate(String fieldName, String fileSuffix, Object value) {
+    public String validate(PipelineInputDefinition pipelineInputDefinition, Object value) {
+      String fieldName = pipelineInputDefinition.getName();
       String stringArrayErrorMessage = "%s must be an array of strings".formatted(fieldName);
       if (value == null) {
         return NOT_NULL_OR_EMPTY_ERROR_MESSAGE.formatted(fieldName);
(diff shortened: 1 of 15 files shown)


service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -165,9 +167,10 @@ public String validate(String fieldName, Object value) {
    * the value fails validation, return a descriptive error message in a String.
    *
    * @param fieldName - the name of the field being validated (used to construct error messages)
+   * @paramf fileSuffix - nullable suffix used to validate file types
    * @param value - the value to validate
    */
-  public abstract String validate(String fieldName, Object value);
+  public abstract String validate(String fieldName, String fileSuffix, Object value);
```
jsotobroad, line 173: hmm i think this signature might be better if we change it to 
```
public abstract String validate(PipelineVariableTypesEnum pipelineVariableTypesEnum, Object value);
```
and then inside the overriding functions we can grab the field name and the suffix if its needed.   What do you think?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -43,24 +43,24 @@ public <T> T cast(String fieldName, Object value, TypeReference<T> typeReference
...
-  VCF {
+  FILE {
     @Override
     public <T> T cast(String fieldName, Object value, TypeReference<T> typeReference) {
       return (T) STRING.cast(fieldName, value, new TypeReference<>() {});
     }
 
     @Override
-    public String validate(String fieldName, Object value) {
+    public String validate(String fieldName, String fileSuffix, Object value) {
```
jsotobroad, line 60: do we think we will have a use case where we allow for multiple suffixes or we dont care about the suffix at all? if so how will this method handle those use cases

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
mmorgantaylor: hm, the suffix isn't stored in the `PipelineVariableTypesEnum` object, it's in each individual `PipelineInputDefinition` object. if we're confident we're only ever going to call validate on inputs, then we could do 
`public abstract String validate(PipelineInputDefinition pipelineInputDefinition, Object value)`
and the rest like you suggest. what do you think?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
mmorgantaylor: great questions. for now I'm ok supporting exactly one suffix (and I should add a test for that in the database test) and adding support for none or multiple if/when we need it. do you have strong feelings otherwise?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
mmorgantaylor: if we do pass in the PipelineInputDefinition directly for now, and later decide we want to run validate on pipeline outputs too, then `fileSuffix` will be added to the `BasePipelineVariableDefinition` and we should be able to amend the signature to take an implementation of `BasePipelineVariableDefinition`, right? i.e. doing this can be compatible with validating on outputs if we decide to do that in the future

jsotobroad: yup i copy pasted the wrong thign lol.  this is exactly right. PipelineInputDefinition for now and if we want to validate outputs (which we _shouldnt_ have to) then BasePipelineVariableDefinition with a file_suffix being added to the outputs table

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
jsotobroad: hmm  yea i would say for the multiple use case not worth thinking about.  I could go either way on the suffix-less use case.  I dont think any of the possible imputation pipeline inputs we get from users would be a suffix-less use case.  If this is true then yea i would say coding for one speicfic output for every FILE input makes sense

mmorgantaylor: cool. technically currently if we have a file input and don't want to check for a suffix, we can have the fileSuffix be an empty string and the validate check will pass. (I just added a test for this)

service/src/test/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnumTest.java
```diff
@@ -0,0 +1,272 @@
...
+    PipelineInputDefinition stringInputDefinition =
+        new PipelineInputDefinition(
+            1L, commonInputName, "string_input", STRING, null, true, true, null);
+    PipelineInputDefinition fileVcfInputDefinition =
+        new PipelineInputDefinition(
+            1L, commonInputName, "file_vcf_input", FILE, ".vcf.gz", true, true, null);
+    PipelineInputDefinition fileBedInputDefinition =
+        new PipelineInputDefinition(
+            1L, commonInputName, "file_bed_input", FILE, ".bed", true, true, null);
+    PipelineInputDefinition fileNoSuffixInputDefinition =
```
mmorgantaylor, line 51: @jsotobroad  here + line 116 is the test I added to show we can define a FILE input without a suffix check by using an empty string as the fileSuffix value.

jsotobroad (approved):
woohoo awesome

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
mmorgantaylor: ok perfect. great idea, thanks!

PR #107, opened 2024-07-29
Add WDL pipelines to Dockstore

mmorgantaylor:
### Description 

In order to import workflows into GCP Terra workspaces, they need to be either in the Methods Repo (Agora) or Dockstore. Dockstore allows automatic updates via github, so that's what we're doing. This PR adds the required .dockstore.yml file to specify the workflows to sync in Dockstore.


jsotobroad (approved):
dockstore......

PR #108, opened 2024-08-01
TSPS-274 add rawls client to codebase

jsotobroad:
### Description 

Rawls does not generate their own client and publish it so this PR creates the rawls client by downloading their openapi yml and generating it ourselves.  This is how it seems other [teams](https://github.com/DataBiosphere/terra-data-catalog/blob/main/rawls-client/swagger.gradle) do it.  It definitely has some downsides that we should chat about but for now thisis the first pass.

This shoudl be merged in after https://github.com/broadinstitute/terra-helmfile/pull/5774


service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsClient.java
```diff
@@ -0,0 +1,40 @@
...
+
+  protected ApiClient getUnauthorizedApiClient() {
+    // Share one api client across requests.
+    return new ApiClient()
+        .setHttpClient(commonHttpClient)
+        .setBasePath(rawlsConfig.baseUri())
+        .setDebugging(rawlsConfig.debugApiLogging());
+  }
+
+  EntitiesApi getEntitiessApi(String accessToken) {
```
mmorgantaylor, line 33: typo (extra 's')

service/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation "org.databiosphere:workspacedataservice-client-okhttp-jakarta:0.2.115-20240227.195850-1"
     implementation "bio.terra:cbas-client:0.0.199"
     implementation "bio.terra:workspace-manager-client:0.254.1124-20240618.154106-1"
+    implementation project(':rawls-client')
```
mmorgantaylor, line 52: not your code, but while you're here could you add a space after `implementation` in line 54?

rawls-client/build.gradle
```diff
@@ -0,0 +1,16 @@
+import org.springframework.boot.gradle.plugin.SpringBootPlugin
+
+plugins {
+    id 'bio.terra.pipelines.java-library-conventions'
+    id 'io.spring.dependency-management'
+    id 'org.hidetake.swagger.generator'
+    id 'de.undercouch.download'
```
mmorgantaylor, line 7: i was curious what this does, looks like it's a progress meter for downloads - is this actually used somewhere? when i run locally i don't see anything that looks like it's using this, just: 

```
> Task :rawls-client:downloadRawlsYaml
Download https://raw.githubusercontent.com/broadinstitute/rawls/874b2e1bc6d95f90e1cf0db7569cfbaae9da34bf/core/src/main/resources/swagger/api-docs.yaml
```

mmorgantaylor (approved):
looks great, one typo and some curiosity questions

rawls-client/build.gradle
jsotobroad: so the download command is using that plugin (if you remove the plugin the download wont be able to work).  I think the reason you arent seeing the progress is because it seems to dissapear after its done - https://github.com/michel-kraemer/gradle-download-task/blob/master/screencast/screencast.svg

PR #110, opened 2024-08-06
TSP-275 Add GCP workspace identifiers to database and admin update endpoint

mmorgantaylor:
### Description 

Since we'll be using GCP workspaces, which are often identified in APIs via workspace_project and workspace_name, we add these fields to the `pipelines` and `pipeline_runs` table, in addition to the existing workspace_id field, which may be used in future as well.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-275


common/openapi.yml
```diff
@@ -749,10 +765,14 @@ components:
       description: |
         json object containing the admin provided information to update a pipeline with
       type: object
-      required: [ workspaceId ]
+      required: [ workspaceProject, workspaceName ]
```
mmorgantaylor, line 768: for now, only workspace project and name are required. workspace id might become required in the future, but currently GCP workspaces don't have user-facing workspace ids.

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -59,16 +59,29 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
   }
 
   /**
-   * This method is meant to only be called by an admin endpoint to update the control workspace id
-   * for a pipeline
+   * This method is meant to only be called by an admin endpoint to update the control workspace id,
+   * workspace project, and/or workspace name for a pipeline. If any of the inputs are null, that
```
mmorgantaylor, line 63: the side effect of this strategy is it's impossible to nullify a populated value via this method. if we think we want this ability, we can change strategy a bit

jsotobroad (commented):
questions around how much we want to deal with workspaceid

common/openapi.yml
```diff
@@ -494,6 +494,10 @@ components:
           $ref: "#/components/schemas/PipelineDescription"
         workspaceId:
```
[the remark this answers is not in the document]
mmorgantaylor: I think we still want to see it here if it is populated

common/openapi.yml
mmorgantaylor: that's a good point. probably none should be required here, since the intent is to allow updating any of them independently.

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -67,8 +67,11 @@ public ResponseEntity<ApiAdminPipeline> updatePipeline(
     PipelinesEnum validatedPipelineName =
         PipelineApiUtils.validatePipelineName(pipelineName, logger);
     UUID workspaceId = body.getWorkspaceId();
```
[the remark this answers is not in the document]
mmorgantaylor: that works for me

common/openapi.yml
mmorgantaylor: instead, we agreed to remove workspaceId from these endpoints, so these two fields should be required after all (here and in the update function)

service/src/main/java/bio/terra/pipelines/db/entities/PipelineRun.java
```diff
@@ -37,9 +37,15 @@ public class PipelineRun {
   @Column(name = "pipeline_id", nullable = false)
   private Long pipelineId;
 
-  @Column(name = "workspace_id", nullable = false)
+  @Column(name = "workspace_id")
   private UUID workspaceId;
 
+  @Column(name = "workspace_project")
```
jsotobroad, line 43: nitpicky but these could be updated to be `nullable = false` for now but would be removed later if we went back to adding workspaceId hahaha so nvm :)

jsotobroad (approved):

PR #112, opened 2024-08-19
TSPS-276 Use rawls to submit/monitor workflow and add wds row

jsotobroad:
### Description 

Still in draft, currently requires manual changes to rawls openapi.yml to work

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-276
https://broadworkbench.atlassian.net/browse/TSPS-277
https://broadworkbench.atlassian.net/browse/TSPS-294

PR #115, opened 2024-08-21
TSPS-300 store control workspace storage container url in db, update via admin api

mmorgantaylor:
### Description 

Rather than hardcoding the control workspace's storage container (bucket) url, we now store it in the db and can update it via an admin API.

We require all the workspace values to be provided in the admin update api to be not null.

Also an add-on update to https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/112 now that we have an "emptyWDL" dev control workspace to use that has the correct WDL method name.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-300

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
@@ -85,7 +85,9 @@ public class PipelineRunsService {
 
     PipelinesEnum pipelineName = pipeline.getName();
 
-    if (pipeline.getWorkspaceProject() == null || pipeline.getWorkspaceName() == null) {
+    if (pipeline.getWorkspaceProject() == null
+        || pipeline.getWorkspaceName() == null
+        || pipeline.getWorkspaceStorageContainerUrl() == null) {
       throw new InternalServerErrorException("%s workspace not defined".formatted(pipelineName));
     }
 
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -64,12 +64,17 @@ public Pipeline getPipeline(PipelinesEnum pipelineName) {
...
-      PipelinesEnum pipelineName, String workspaceProject, String workspaceName) {
+      PipelinesEnum pipelineName,
+      String workspaceProject,
+      String workspaceName,
+      String workspaceStorageContainerUrl) {
     Pipeline pipeline = getPipeline(pipelineName);
     pipeline.setWorkspaceProject(workspaceProject);
     pipeline.setWorkspaceName(workspaceName);
+    pipeline.setWorkspaceStorageContainerUrl(workspaceStorageContainerUrl);
     pipelinesRepository.save(pipeline);
```
jsotobroad, line 78: just double checking, because we require all the fields in the request body, none of the values passed in should be null right?  curious if for this we want to add the @NotNull and @Validated annotations since this is a super important operation for our service - https://www.baeldung.com/java-notnull-method-parameter#1-spring-boot-validation

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -95,19 +97,14 @@ public void preparePipelineRun(
               .formatted(jobId));
     }
 
-    // hardcode for now, ultimately fetch from Rawls or store in config?
-    // this is for dev control workspace:
-    // https://bvdp-saturn-dev.appspot.com/#workspaces/teaspoons-imputation-dev/teaspoons_imputation_dev_control_workspace_20240726
-    String workspaceStorageContainerUrl = "gs://fc-secure-972057b6-fce1-4ec9-be0c-0e5c97fdc3e8";
```
jsotobroad, line 101: Im a dummy and forgot to do this in my PR but since you already created the empty beagle workspace, do you mind updating the SubmitCromwellSubmissionStep to use the `wdlMethodName` when creating the submission.

jsotobroad (commented):
1 question and silly ask.  looks good though

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: good idea, thanks!

jsotobroad (approved):
woohoo thanks!

PR #121, opened 2024-09-09
TSPS-303 validate method config and update if needed

jsotobroad:
### Description 
We want to make sure we are using the correct wdl and inputs when submitting our pipelines.  This PR adds the logic to the submission step in the flight make this possible.

Do we think this is too much to do in one step?

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-303

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
@@ -155,7 +155,10 @@ public class RawlsService implements HealthCheck {
   }
 
   /**
-   * validates a method config against the expected version, inputs, and outputs
+   * validates a method config against the expected version, inputs, and outputs it does this by
+   * checking that the methodRepoMethod method version matches we expect and that all expected
+   * inputs outputs both exist in the method config's list of inputs/outputs and that the data table
+   * reference for each input/output are what we expect.
    *
    * @param methodConfiguration - method config to validate against
    * @param wdlWorkflowName - name of the wdl workflow, used to construct the full wdl variable name
@@ -205,11 +208,14 @@ public class RawlsService implements HealthCheck {
 
   /**
    * this method takes a pre-existing method configuration and updates it to match what the service
-   * expects. It does this by updating the methodRepoMethod methodVersion and by setting the inputs
-   * and outputs fields to match the references we expect.
+   * expects. It does this by setting the input and outputs fields to match what we expect both from
+   * what keys we expect and the data table refrence we expect.
+   *
+   * <p>It also sets the methodRepoMethod method version to the version we should be running as well
+   * as updating the methodRepoMethod method uri by swapping the old version string with the
+   * expected version string.
    *
    * @param methodConfiguration - method configuration to update
-   * @param methodConfiguration - method config to validate against
    * @param wdlWorkflowName - name of the wdl workflow, used to construct the full wdl variable name
    * @param pipelineInputDefinitions - list of input definitions for a pipeline
    * @param pipelineOutputDefinitions - list of output definitions for a pipeline
@@ -223,7 +229,7 @@ public class RawlsService implements HealthCheck {
       List<PipelineInputDefinition> pipelineInputDefinitions,
       List<PipelineOutputDefinition> pipelineOutputDefinitions,
       String wdlMethodVersion) {
-    // update wdl method version and uri to match version
+    // set wdl method version and uri to match the expected version
     String oldVersion = methodConfiguration.getMethodRepoMethod().getMethodVersion();
     String oldUri = methodConfiguration.getMethodRepoMethod().getMethodUri();
     methodConfiguration.getMethodRepoMethod().setMethodVersion(wdlMethodVersion);
@@ -231,7 +237,7 @@ public class RawlsService implements HealthCheck {
         .getMethodRepoMethod()
         .setMethodUri(oldUri.replace(oldVersion, wdlMethodVersion));
 
-    // update inputs
+    // set inputs
     HashMap<String, String> expectedInputs = new HashMap<>();
     for (PipelineInputDefinition pipelineInputDefinition : pipelineInputDefinitions) {
       String fullWdlVariableName =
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -94,11 +105,140 @@ public Entity upsertDataTableEntity(
...
+                    methodName));
+  }
+
   // returns true if submission is in a running state
   public static boolean submissionIsRunning(Submission submission) {
     return !FINAL_RUN_STATES.contains(submission.getStatus());
   }
 
+  /**
+   * validates a method config against the expected version, inputs, and outputs
```
mmorgantaylor, line 144: could use a little more description here - i like that you validate not only that the expected inputs and outputs are there, but also that they're pointed to the corresponding fields in the data table. if you could note that here that'd be great.

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
```diff
@@ -59,10 +71,65 @@ public StepResult doStep(FlightContext flightContext) {
...
+    boolean validMethodConfig =
+        rawlsService.validateMethodConfig(
+            methodConfiguration,
+            wdlMethodName,
+            inputDefinitions,
+            outputDefinitions,
+            wdlMethodVersion);
+
+    // if not a valid method config, set the method config to what we think it should be.  This
+    // shouldn't happen
```
mmorgantaylor, line 103: shouldn't happen often, but it will happen once every time we change the wdlMethodVersion... unless we update it in Terra as well as in our service, which actually we could do. 🤷 it's nice to have the warning log here. wonder if we should set up a slack channel to capture warnings.

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
```diff
@@ -59,10 +71,65 @@ public StepResult doStep(FlightContext flightContext) {
...
+    List<PipelineOutputDefinition> outputDefinitions =
+        inputParameters.get(
+            RunImputationJobFlightMapKeys.PIPELINE_OUTPUT_DEFINITIONS, new TypeReference<>() {});
 
     // validate and extract parameters from working map
     FlightMap workingMap = flightContext.getWorkingMap();
 
+    // grab current method config and validate it
+    MethodConfiguration methodConfiguration =
+        rawlsService.getCurrentMethodConfigForMethod(
```
mmorgantaylor, line 89: do we want to wrap this in try-catch retry logic like the later rawls call?

mmorgantaylor (commented):
looks good to me so far!

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
jsotobroad: is that something we can do -- is that a sentry thingy?

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
jsotobroad: oh good point, definitely missed this.

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -108,11 +120,150 @@ public Entity getDataTableEntity(
...
+                    methodName));
+  }
+
   // returns true if submission is in a running state
   public static boolean submissionIsRunning(Submission submission) {
     return !FINAL_RUN_STATES.contains(submission.getStatus());
   }
 
+  /**
+   * validates a method config against the expected version, inputs, and outputs it does this by
```
mmorgantaylor, line 159: nit: missing period after "outputs"

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -108,11 +120,150 @@ public Entity getDataTableEntity(
...
+   * checking that the methodRepoMethod method version matches we expect and that all expected
```
mmorgantaylor, line 160: - "methodRepoMethod method version" can just be "wdlMethodVersion" 
- "matches **what** we expect"
- (next line) "inputs outputs both exist" -> "inputs/outputs exist"

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
mmorgantaylor: yeah probably sentry

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
```diff
@@ -59,10 +71,71 @@ public StepResult doStep(FlightContext flightContext) {
...
+            wdlMethodName,
+            inputDefinitions,
+            outputDefinitions,
+            wdlMethodVersion);
+
+    // if not a valid method config, set the method config to what we think it should be.  This
+    // shouldn't happen
+    if (!validMethodConfig) {
+      logger.warn(
+          "found method config that was not valid for billing project: {}, workspace: {}, method name: {}, methodConfigVersion: {}",
```
mmorgantaylor, line 112: nit: it's not necessarily "not valid" but it's not matching our expectations. super minor but to me "not valid" suggests there's an actual error in the setup

mmorgantaylor (commented):
mostly minor comments, one question about validating the root entity in the method config

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/gcp/SubmitCromwellSubmissionStep.java
jsotobroad: but to us there is an error in the setup no?

mmorgantaylor (approved):
looks awesome!!

PR #133, opened 2024-09-20
update wdlMethodVersion regex check to work with warp tag patterns

jsotobroad:
### Description 

we want to update our wdlmethodversion check so that we can use wdls released in warp

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
@@ -112,13 +112,15 @@ public class PipelinesService {
     pipeline.setWorkspaceStorageContainerName(workspaceStorageContainerUrl);
     pipeline.setWorkspaceGoogleProject(workspaceGoogleProject);
 
-    // ensure wdlMethodVersion follows semantic versioning regex (only numbers allowed)
+    // ensure wdlMethodVersion follows semantic versioning regex (can be preceded by a string ending
+    // in v)
     final Pattern pattern = Pattern.compile(SEM_VER_REGEX_STRING);
     final Matcher matcher = pattern.matcher(wdlMethodVersion);
     if (!matcher.matches()) {
       throw new ValidationException(
           String.format(
-              "wdlMethodVersion %s does not follow semantic versioning regex", wdlMethodVersion));
+              "wdlMethodVersion %s does not follow semantic versioning regex %s",
+              wdlMethodVersion, SEM_VER_REGEX_STRING));
     }
 
     // ensure that major version of wdlMethodVersion matches the value of the pipeline version.
--- a/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
@@ -49,7 +49,7 @@ public class PipelinesService {
   private final SamService samService;
 
   private static final String SEM_VER_REGEX_STRING =
-      "^(0|[1-9]\\d*|[a-zA-Z_]*v[0-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$";
+      "^(0|[1-9]\\d*|[a-zA-Z_]*v\\d+)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$";
 
   @Autowired
   public PipelinesService(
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -227,6 +228,13 @@ void updatePipelineWorkspace() {
     // assert the fetched info from Rawls has been written
     assertEquals(newWorkspaceStorageContainerName, p.getWorkspaceStorageContainerName());
     assertEquals(newWorkspaceGoogleProject, p.getWorkspaceGoogleProject());
+
+    // update pipeline with warp-like wdlMethodVersion
+    pipelinesService.updatePipelineWorkspace(
```
mmorgantaylor, line 233: should we assert anything after this? i get that we just expect this to work without throwing an error, but it could theoretically silently fail somehow?

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -235,7 +243,10 @@ private static Stream<Arguments> badWdlMethodVersions() {
         arguments("blah.3.2"),
         arguments("0.2"),
         arguments("1.bhmm.2"),
-        arguments("1.4.ok"));
+        arguments("1.4.ok"),
+        arguments(
+            "ImputationBeagle_development_0.0.0"), // missing a "v" before the semantic version
+        arguments("hiiv.1.4"));
```
mmorgantaylor, line 249: do we expect "v0.1.2" to work? can we put it in a test?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -49,7 +49,7 @@ public class PipelinesService {
   private final SamService samService;
 
   private static final String SEM_VER_REGEX_STRING =
-      "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$";
+      "^(0|[1-9]\\d*|[a-zA-Z_]*v\\d+)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$";
```
mmorgantaylor, line 52: since the "v" is so important, should we add a note about formatting somewhere, like maybe in openapi.yml at the wdlMethodVersion description?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -122,7 +122,10 @@ public Pipeline updatePipelineWorkspace(
     }
 
     // ensure that major version of wdlMethodVersion matches the value of the pipeline version.
-    if (pipeline.getVersion().equals(Integer.parseInt(wdlMethodVersion.split("\\.")[0]))) {
+    // split wdlMethodVersion by 'v' and take the last element of the resulting array
+    String splitWdlMethodVersion =
```
mmorgantaylor, line 126: minor: can we call this "wdlMethodExtractedSemVer" or something? "split" suggested to me that it was the array of strings after the split, not the last element in the array

mmorgantaylor (commented):
looks great, a few minor questions

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
jsotobroad: im not sure we want to list all the possible options in the openapi yaml, its not like `v` is required.  Shoudl we put the regex in the openapi yml description?

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
jsotobroad: we do expect it to work, its part of the same regex group as the current test.  do you want one specifically with this string?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
jsotobroad: i added the regex to the error message

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
jsotobroad: im not sure how it could silently fail here.  do you mean like the repository doesn't save anything even though it says it did?

service/src/test/java/bio/terra/pipelines/service/PipelinesServiceTest.java
```diff
@@ -251,6 +258,27 @@ void updatePipelineWorkspaceBadWdlMethodVersion(String badWdlMethodVersion) {
                 pipelinesEnum, newWorkspaceBillingProject, newWorkspaceName, badWdlMethodVersion));
   }
 
+  private static Stream<Arguments> goodWdlMethodVersions() {
```
mmorgantaylor, line 261: thank you so much for adding this, this helps me a ton

mmorgantaylor (approved):
awesome, thank you for adding the clear tests and error response.

PR #135, opened 2024-09-23
[TSPS-322] make explicit storage container url in PrepareImputationInputsStep

mmorgantaylor:
### Description 

The previous conversion from storageContainerUrl to storageContainerName caused a problem in PreparePipelineInputsStep since it expects that value to be the full Url. Here we add a value to the flight working map to specify the protocol to append to the storageContainerName.

Also did a refactor of a bunch of methods related to input and output formatting and generation into a new PipelineInputsOutputsService.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-332

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
@@ -23,16 +23,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 /* Service to encapsulate the logic for processing pipeline inputs and outputs */
 @Service
 public class PipelineInputsOutputsService {
-  private static final Logger logger = LoggerFactory.getLogger(PipelineInputsOutputsService.class);
-
   private final GcsService gcsService;
 
   private final PipelineInputsRepository pipelineInputsRepository;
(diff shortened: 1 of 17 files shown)


service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
```diff
@@ -0,0 +1,199 @@
...
+import org.springframework.stereotype.Service;
+
+/* Service to encapsulate the logic for processing pipeline inputs and outputs */
+@Service
+public class PipelineInputsOutputsService {
+  private final GcsService gcsService;
+
+  private final PipelineInputsRepository pipelineInputsRepository;
+  private final PipelineOutputsRepository pipelineOutputsRepository;
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
jsotobroad, line 36: so i think we want to autowire this in so we can use the one from the bean we create as opposed to the default one (we should do this everywhere we dont already do this)

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightMapKeys.java
```diff
@@ -8,6 +8,8 @@ public abstract class RunImputationJobFlightMapKeys {
   public static final String ALL_PIPELINE_INPUTS = "all_pipeline_inputs";
   public static final String CONTROL_WORKSPACE_STORAGE_CONTAINER_NAME =
       "control_workspace_storage_container_name";
+  public static final String CONTROL_WORKSPACE_STORAGE_CONTAINER_PROTOCOL =
+      "control_workspace_storage_container_protocol";
```
jsotobroad, line 12: i wonder what the limit of these strings are in the stairway database hahaha

jsotobroad (commented):
This looks so much cleaner! One comment about object mapper usage

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightMapKeys.java
mmorgantaylor: haha yeah we apparently haven't hit the limit yet. want me to find out?

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
mmorgantaylor: ah thanks, forgot about that

jsotobroad (approved):

PR #136, opened 2024-09-24
TSPS-326 add wdl to mask or subset a vcf given a bed file

jsotobroad:
### Description 

We want to use this wdl for scientific validation of the refernece panel.  We will mask wgs samples down to an array build's sites and then impute and we shoudl get something close to the original wgs sample

Example of workflow running - https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/tsps_gcp_scratch_space_mma/job_history/dd9e56bd-63d6-43d4-bd9f-93caf91dcdae

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-326

### DIFF
--- a/pipelines/imputation/scientificValidation/README.md
+++ b/pipelines/imputation/scientificValidation/README.md
@@ -18,7 +18,12 @@ algorithm has been run
 ## SubsetVcfByBedFile
 ### Purpose
 This wdl is meant to be used to subset a vcf down
-to sites provided through a bed file
+to sites provided through a bed file.  This wdl does
+not interact with headers or annotations mostly because
+the only really "required" header is the dictionary
+and that gets transferred across and the imputation
+tool only look at GT and no info/format fields so
+we can just leave them be and have it not affect
 
 #### Inputs
 * input_vcf - input file to be subset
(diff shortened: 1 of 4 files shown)


pipelines/imputation/scientificValidation/README.md
```diff
@@ -13,3 +13,18 @@ This wdl is basically a wrapper around that tool/image
 #### Outputs
 * recombined_reference_panel - output vcf after mitigation 
 algorithm has been run
+
+
+## ReshapeReferencePanel
```
mmorgantaylor, line 18:
```suggestion
## SubsetVcfByBedFile
```

mmorgantaylor (commented):
Is the only thing we care about that the proper sites are filtered down? do we care about headers or annotations or anything like that that are expected for an array but wouldn't be present in a WGS vcf?

have you run the output of this through imputation?

jsotobroad:
> Is the only thing we care about that the proper sites are filtered down? do we care about headers or annotations or anything like that that are expected for an array but wouldn't be present in a WGS vcf?
> 
> have you run the output of this through imputation?

Outside of the dictionary lines im not sure if there are any other significant headers that we should care about

mmorgantaylor (approved):
please add a couple lines in the readme to note that we don't care about headers/annotations because the imputation tools don't care about those either.

PR #139, opened 2024-09-30
TSPS-322 Add functioning b2c login to CLI

nmalfroy:
### Description 

Can now build the thick CLI, and it handles auth to Terra b2c.

Corresponding terraform-ap-deployments PR to register the local redirect URL with b2c: https://github.com/broadinstitute/terraform-ap-deployments/pull/1716

Note that tests are not included here but will be implemented in TSPS-321, as will packaging and publishing the cli.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-322


common/openapi.yml
```diff
@@ -543,12 +543,14 @@ components:
             $ref: '#/components/schemas/JobReport'
 
     GetPipelinesResult:
-      type: array
-      items:
-        $ref: '#/components/schemas/Pipeline'
+      description: result of a getPipelines request
+      type: object
```
mmorgantaylor, line 547: changing from object to array will allow this response to be processed properly by pydantic, we think

teaspoons-cli/teaspoons/commands/pipelines.py
```diff
@@ -0,0 +1,36 @@
...
+        click.echo(str(e), err=True)
+        exit(1)
+
+@pipelines_app.command()
+@click.argument('name')
+def get_info(name: str):
+    try:
+        with ClientWrapper() as api_client:
+            pipeline_client = PipelinesApi(api_client=api_client)
+            pipeline = pipeline_client.get_pipeline_details(pipeline_name=name)
```
jsotobroad, line 32: just cuz im dumb. Is the idea that if we wanted to do some more fancy logic other than just make an api call we would do like `pipelines.py` -> `service/pipelines_service.py` -> `pipeline_client.py` (ignore the names i used hahaha) or something liek that where the bulk of the logic would be in that middle layer?

teaspoons-cli/.env
```diff
@@ -0,0 +1,10 @@
+# Teaspoons API URL
+TEASPOONS_API_URL=https://tsps.dsde-dev.broadinstitute.org
```
jsotobroad, line 2: i assume at some point we'll want to parameterized this so it can run on different environments

teaspoons-cli/teaspoons/config.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self):
+        self.config = dotenv_values("../.env")
+
+        self.client_info = OAuth2ClientInfo.from_oidc_endpoint(
+            self.config["OAUTH_OPENID_CONFIGURATION_URI"],
+            client_id=self.config["OAUTH_CLIENT_ID"],
+            scopes=[f"openid+email+profile+{self.config['OAUTH_CLIENT_ID']}"]
+        )
+
+        # Figure out the path to the token file...there must be some way to abstract out local storage
```
jsotobroad, line 20: what does this comment mean about abstracting local storage?  like not needing to save the "path" in the config file or something else?

teaspoons-cli/teaspoons/commands/auth.py
```diff
@@ -0,0 +1,24 @@
...
+
+auth_app = typer.Typer()
+
+@auth_app.command()
+def login():
+    token = _load_local_token(cli_config.token_file)
+    if token and _validate_token(token):
+        click.echo('Already authenticated')
+        return
+    token = get_access_token_with_browser_open(cli_config.client_info, server_port=10444)
```
jsotobroad, line 18: should the port be a config value?

teaspoons-cli/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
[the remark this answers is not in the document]
mmorgantaylor: don't think so, though probably update the year...

teaspoons-cli/README.md
```diff
@@ -1,6 +1,72 @@
...
+├── teaspoons
+│   └── commands
+│   │   └── __init__.py
+│   │   └── auth.py
+│   │   └── pipelines.py
+│   └── generated
+│   │   └── [auto-generated files for thin client]
+├── └── tests (to be created)
+│   └── __init__.py
+│   └── auth.py
```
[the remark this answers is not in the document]
mmorgantaylor: they do different things - i like your suggestion below to name this one auth_helper.py

teaspoons-cli/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "teaspoons-cli"
+version = "0.0.1"
+description = "Command line interface for interacting with the Terra Scientific Pipelines Service, or Teaspoons."
+authors = ["Terra Scientific Services <teaspoons-developers@broadinstitute.org>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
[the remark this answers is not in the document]
mmorgantaylor: so this means we support 3.9 and up (that's what the `^` means). 3.9 isn't EOL till late 2025. i don't have strong feelings about this, happy to bump to 3.10 and up if that sounds better to you?

teaspoons-cli/teaspoons/auth.py
```diff
@@ -0,0 +1,90 @@
+
```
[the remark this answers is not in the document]
mmorgantaylor: great suggestion, thanks!

teaspoons-cli/teaspoons/client.py
```diff
@@ -0,0 +1,27 @@
...
+
+cli_config = CliConfig()  # initialize the config from environment variables
+
+def _get_api_client(token: str) -> ApiClient:
+    api_config = Configuration()
+    api_config.host = cli_config.config["TEASPOONS_API_URL"]
+    api_config.access_token = token
+    return ApiClient(configuration=api_config)
+
+class ClientWrapper:
```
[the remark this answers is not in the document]
mmorgantaylor: it's convention to use "CapCase" for python classes: https://peps.python.org/pep-0008/#class-names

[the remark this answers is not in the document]
mmorgantaylor: oh do you mean the file name? that's ok with me if you'd prefer it, though this script contains more than just the ClientWrapper class

teaspoons-cli/pyproject.toml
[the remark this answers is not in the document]
jsotobroad: ahhh gotcha, TIL.  yea that seems good. I thought it meant we just support 3.9.  I'll try installing it on 3.12 or w/e and see what happens

teaspoons-cli/teaspoons/client.py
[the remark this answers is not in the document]
jsotobroad: thats fair, i take back my comment!  (and yea was referring to the file name)

teaspoons-cli/teaspoons/commands/auth.py
mmorgantaylor: good idea

teaspoons-cli/teaspoons/commands/pipelines.py
mmorgantaylor: yes, that's my understanding too - there's currently a missing layer for any logic that's not a straight wire-through of the thin client, since we haven't implemented any of those yet.

teaspoons-cli/teaspoons/config.py
mmorgantaylor: honestly not sure, Muscles left this comment. I think he was hoping there'd be a more elegant way to do the following if-else statement, but what he has makes sense to me. should i remove this comment (or the second half of it)?

jsotobroad: im trying to think when we would ever set the `LOCAL_STORAGE_PATH` to start with a `/` 🤔.  we can probably take out the if part of the if/else and add it back in if we ever feel like its necessary?

teaspoons-cli/.env
mmorgantaylor: yes, added to AC for TSPS-321

jsotobroad (approved):

teaspoons-cli/teaspoons/commands/pipelines.py
```diff
@@ -0,0 +1,32 @@
...
+# teaspoons modules
+from client import ClientWrapper
+from utils import _pretty_print, handle_api_exceptions
+
+pipelines_app = typer.Typer()
+
+
+@pipelines_app.command()
+@handle_api_exceptions
+def list():
```
mmorgantaylor, line 18: note this won't work (while pointing to dev teaspoons) until this PR is merged.

davidangb (commented):
I didn't review at the syntax level. At the design level, this looks solid to me. Some questions to think about:
* will you be publishing the client to pypi or elsewhere?
* If I understand the code correctly, the "thin client" and "thick client" are just different files/directories. Would you consider generating, versioning, and publishing the thin client as its own package, and then importing that package into the thick client? That could help enforce isolation between the two.

teaspoons-cli/teaspoons/config.py
```diff
@@ -0,0 +1,22 @@
...
+class CliConfig:
+    """A class to hold configuration information for the CLI"""
+
+    def __init__(self):
+        self.config = dotenv_values("../.env")
+
+        self.client_info = OAuth2ClientInfo.from_oidc_endpoint(
+            self.config["OAUTH_OPENID_CONFIGURATION_URI"],
+            client_id=self.config["OAUTH_CLIENT_ID"],
+            scopes=[f"openid+email+profile+{self.config['OAUTH_CLIENT_ID']}"]
```
rtitle, line 17: Does the client ID truly need to be passed as a scope, or does it work with just `openid+email+profile`?

teaspoons-cli/teaspoons/auth_helper.py
```diff
@@ -0,0 +1,94 @@
...
+
+from config import CliConfig
+
+
+cli_config = CliConfig()  # initialize the config from environment variables
+
+
+def get_auth_url(client_info: OAuth2ClientInfo, redirect_uri: str) -> str:
+    """
+    Note: this is overridden from then oauth2-cli-auth library to be able to specify a custom auth url
```
rtitle, line 18: How does this differ from `oauth2_cli_auth.get_auth_url`? Looks like just the addition of `&prompt=login`?

Wonder if you could call the library method and just append the extra `&prompt=login` field..

rtitle (commented):
I didn't test it, but overall the approach looks good to me!

rtitle (approved):

teaspoons-cli/teaspoons/config.py
rtitle: Oh okay, good to know that's still a requirement. It always felt weird to me.

I think the `openid email profile` scopes are configured in the B2C claims provider ([google](https://github.com/broadinstitute/terraform-ap-deployments/blob/master/azure/b2c/policies/TrustFrameworkExtensions.xml.tftpl#L127) and [azure](https://github.com/broadinstitute/terraform-ap-deployments/blob/master/azure/b2c/policies/TrustFrameworkExtensions.xml.tftpl#L228)). So maybe that's why just specifying the client ID here works. I think leaving it as-is is best though. 👍

PR #141, opened 2024-10-03
TSPS-181 Add stairway hooks for failed flights - update metrics and status

mmorgantaylor:
### Description 

We have two actions we want to take if a pipeline run flight fails:
1. Set the pipelineRun status in our db to FAILED
2. Increment the pipeline failed metric counter for that pipeline type

We had put these into the undo step of the first step of a flight, but this was hacky and wouldn't get executed in cases of dismal failures. The better thing to do is to create Stairway hooks that are called at the end of each flight.

Because these hooks rely on specific fields being present in the working map, they would not necessarily work with any arbitrary flight. Therefore, we add logic to these two hooks to check for the presence of an input parameter that specifies whether to execute the task in the hook. 

This PR also includes a refactor of `JobMapKeys` from an enum to a class so it behaves more like the (newly named) `ImputationJobMapKeys` (we weren't gaining anything from its being an enum).

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-181


service/IMPLEMENTATION_NOTES.md
```diff
@@ -22,3 +22,18 @@ These methods return a list of error messages, if any, that are accumulated into
...
+status in the final step of the stairway flight. In the case of a failed pipelineRun, we update its status 
+using a Stairway hook ([StairwaySetPipelineRunStatusHook](src/main/java/bio/terra/pipelines/common/utils/StairwaySetPipelineRunStatusHook.java)).
+
+We use a Stairway hook so that the pipelineRun will be marked as failed regardless of when in the flight the 
+failure occurs, and whether it's a roll-back-able error or a dismal failure.
+
+Similarly, for metrics reporting, we use a Stairway hook ([StairwayFailedMetricsCounterHook](src/main/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHook.java)) 
+to increment the failed metrics counter when a pipelineRun fails.
+
+Because Stairway hooks are applied at the Stairway instance level and not per-flight, we check that the flight
```
mmorgantaylor, line 38: update this

service/src/main/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHook.java
```diff
@@ -0,0 +1,47 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightStatus;
+import bio.terra.stairway.HookAction;
+import bio.terra.stairway.StairwayHook;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * A {@link StairwayHook} that logs a pipeline failure to Prometheus metrics upon flight failure.
```
jsotobroad, line 18: can we add some more documentation around what key to use to enable this functionality on a flight

service/src/main/java/bio/terra/pipelines/common/utils/StairwayLoggingHook.java
```diff
@@ -1,70 +0,0 @@
-package bio.terra.pipelines.common.utils;
```
jsotobroad, line 1: awesome, cleaning up stuff!!

service/src/main/java/bio/terra/pipelines/common/utils/StairwaySetPipelineRunStatusHook.java
```diff
@@ -0,0 +1,50 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightStatus;
+import bio.terra.stairway.HookAction;
+import bio.terra.stairway.StairwayHook;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/** A {@link StairwayHook} that updates the PipelineRun status to FAILED on flight failure. */
```
jsotobroad, line 18: same here about adding some notes about what key(s) the hook expects

service/src/main/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHook.java
```diff
@@ -0,0 +1,47 @@
...
+  public HookAction endFlight(FlightContext context) {
+    if (isContextInvalid(context)) {
+      return HookAction.CONTINUE;
+    }
+
+    if (inputParametersContainTrue(
+            context.getInputParameters(),
+            RunImputationJobFlightMapKeys.DO_INCREMENT_METRICS_FAILED_COUNTER_HOOK)
+        && context.getFlightStatus() != FlightStatus.SUCCESS) {
+      logger.info("Flight failed, incrementing failed flight counter");
```
jsotobroad, line 35: i think we may want to move this log into the `MetricsUtils.incrementPipelineRunFailed` call unless we wnat to put somethign specific about the flight context in there

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightMapKeys.java
```diff
@@ -14,6 +14,12 @@ public abstract class RunImputationJobFlightMapKeys {
   public static final String WDL_METHOD_VERSION = "wdl_method_version";
   public static final String PIPELINE_RUN_OUTPUTS = "pipeline_run_outputs";
 
+  // keys to determine which Stairway hooks to run
+  public static final String DO_SET_PIPELINE_RUN_STATUS_FAILED_HOOK =
```
jsotobroad, line 18: i would consider moving these flight wide keys to their own class.  Also do you know why these keys are `final String`s and the other key class is an enum?

service/src/main/java/bio/terra/pipelines/stairway/imputation/steps/PrepareImputationInputsStep.java
```diff
@@ -133,25 +127,6 @@ public StepResult doStep(FlightContext flightContext) {
 
   @Override
   public StepResult undoStep(FlightContext flightContext) {
-    // this is the first step in RunImputationGcpJobFlight and RunImputationAzureJobFlight.
```
jsotobroad, line 136: yay clean steps!

service/src/test/java/bio/terra/pipelines/common/utils/FlightUtilsTest.java
```diff
@@ -225,4 +226,42 @@ void flightComplete_running_isNotComplete() {
...
+    context.workingMap(null);
+    assertTrue(FlightUtils.isContextInvalid(context));
+
+    // context with working map is not invalid
+    TestFlightContext contextWithWorkingMap = new TestFlightContext();
+    assertFalse(FlightUtils.isContextInvalid(contextWithWorkingMap));
+  }
+
+  @Test
+  void inputParametersContainTrue_true() {
```
jsotobroad, line 246: i am not loving the mix of camel and snake case.  I think you its just as understandable as just `inputParametersContainsTrue` and the one after this as `inputParametersContainsFalse` and `inputParametersContainsNoKey` or just probably better yet just one function that tests all three scenarios since they're not very complicated with a comment or two

service/src/test/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHookTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @BeforeEach
+  void setup() {
+    meterRegistry = new SimpleMeterRegistry();
+    Metrics.globalRegistry.add(meterRegistry);
+    meterRegistry.clear();
+  }
+
+  @Test
+  void endFlight_notPipelineRunTypeFlight_success() throws InterruptedException {
```
jsotobroad, line 34: same comment about the camel casing and snake casing and smooshing these tests down into a smaller number.  These are a little more complicated but with some refactoring it can probably be made more manageable.  `@ParameterizedTest` might be useful here?

service/src/test/java/bio/terra/pipelines/common/utils/StairwaySetPipelineRunStatusHookTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private final UUID testJobId = TestUtils.TEST_NEW_UUID;
+
+  @BeforeEach
+  void setup() {
+    stairwaySetPipelineRunStatusHook = new StairwaySetPipelineRunStatusHook(pipelineRunsService);
+  }
+
+  @Test
+  void endFlight_notPipelineRunTypeFlight_success() throws InterruptedException {
```
jsotobroad, line 36: similar nitpick 🕊️

service/src/main/java/bio/terra/pipelines/common/utils/FlightUtils.java
```diff
@@ -158,4 +158,19 @@ public static boolean flightComplete(FlightState flightState) {
...
+      return true;
+    }
+
+    return false;
+  }
+
+  /** Check whether the inputParameters contain a particular key and whether its value is true */
+  public static boolean inputParametersContainTrue(FlightMap inputParameters, String key) {
+    return inputParameters.containsKey(key)
+        && Boolean.TRUE.equals(inputParameters.get(key, boolean.class));
```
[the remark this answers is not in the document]
mmorgantaylor: yeah, the way Stairway pulls things out, you can't get just a `boolean`. I did it this way to try to ward you off from complaining about Boolean.TRUE.equals, but you're right it should be Boolean.class.

service/src/main/java/bio/terra/pipelines/common/utils/FlightUtils.java
```diff
@@ -158,4 +158,19 @@ public static boolean flightComplete(FlightState flightState) {
         || flightState.getFlightStatus() == FlightStatus.FATAL
         || flightState.getFlightStatus() == FlightStatus.SUCCESS);
   }
+
+  public static boolean isContextInvalid(FlightContext context) {
+    if (context == null || context.getWorkingMap() == null) {
```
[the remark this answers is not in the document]
mmorgantaylor: i just copied this from WSM's hook implementation. i assumed it helped in testing scenarios when you don't want to bother with this, but i didn't test without it - just now did, and everything runs fine, so we can take it out for now.

service/src/main/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHook.java
mmorgantaylor: hmm, MetricsUtils currently doesn't have a logger at all, do you think we should add it?

mmorgantaylor: actually I meant to log the actual status (the way we do in the other hook) - ok to keep this log here in that case?

service/src/main/java/bio/terra/pipelines/stairway/imputation/RunImputationJobFlightMapKeys.java
jsotobroad: yea i dont really have a preference as to which id just rather be consistent if we can. I just noticed that we have two ways of doing it and didnt know if there was a good reason for it

service/src/main/java/bio/terra/pipelines/common/utils/StairwayFailedMetricsCounterHook.java
mmorgantaylor: good catch that it wasn't doing anything with the context!

jsotobroad (approved):
awesome thanks for the updates!

service/src/main/java/bio/terra/pipelines/common/utils/FlightUtils.java
```diff
@@ -158,4 +158,10 @@ public static boolean flightComplete(FlightState flightState) {
         || flightState.getFlightStatus() == FlightStatus.FATAL
         || flightState.getFlightStatus() == FlightStatus.SUCCESS);
   }
+
+  /** Check whether the inputParameters contain a particular key and whether its value is true */
+  public static boolean inputParametersContainTrue(FlightMap inputParameters, String key) {
+    return inputParameters.containsKey(key)
+        && Boolean.TRUE.equals(inputParameters.get(key, Boolean.class));
```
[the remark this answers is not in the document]
snf2ye: Although, sometime stairway gets picky about actually defining variables. Something like this might work instead:
```suggestion
    Boolean booleanKey = inputParameters.get(key, Boolean.class);
    return inputParameters.containsKey(key) && booleanKey;
```
You wouldn't think that pre-defining it in a variable would help, but I've come across a few cases where it is needed.

service/src/main/java/bio/terra/pipelines/common/utils/StairwayLoggingHook.java
snf2ye, line 1: Why is this getting removed?

snf2ye: Ah, I see - it has been moved to [TCL](https://github.com/DataBiosphere/terra-common-lib/blob/6f12b29c98ef45c1f295aa365066fde9ac7bc58a/src/main/java/bio/terra/common/stairway/StairwayLoggingHook.java#L10).

snf2ye (approved):
A few more questions, but LGTM

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
```diff
@@ -3,28 +3,28 @@
 import java.util.Arrays;
 import java.util.List;
 
-public enum JobMapKeys {
+public class JobMapKeys {
```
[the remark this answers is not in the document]
mmorgantaylor: I don't think this is any new standard - when we copied things from WSM, they had the enum JobMapKeys and a class for custom keys. We realized that there wasn't really a reason to have both (and it was confusing to have both), and the class version is less bulky, like you said. so we're going for it!

service/src/main/java/bio/terra/pipelines/common/utils/FlightUtils.java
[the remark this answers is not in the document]
mmorgantaylor: ugh yeah this is super annoying, I don't know if TDR uses SonarQube, but it complains about this approach: 
image
because a `Boolean` can be null. I couldn't figure out how to extract a `boolean` from the inputParameters.

I think that we could get around this by saying if it's null, assume that's false... I'll do some poking... we hate the Boolean.TRUE.equals() thing too 😭

PR #147, opened 2024-10-24
TSPS-321 cli unit tests

mmorgantaylor:
### Description 

Here we refactor the CLI code a bit and clean up the tools used, and add tests.

Future PRs (other tickets) will move this code to its own repo and handle publishing of the thin and thick clients.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-321


teaspoons-cli/teaspoons/auth_helper.py
```diff
@@ -42,7 +49,7 @@ def _validate_token(token: str) -> bool:
         # This is just to ensure the token is not expired
         jwt.decode(token, options={"verify_signature": False, "verify_exp": True})
         return True
-    except Exception as e:
+    except Exception:
```
fedorgrab, line 52: I think it could improve error handling to catch more specific exceptions here. Like this:
```
except jwt.ExpiredSignatureError:
    # log message that token is expired
    return False
except Exception as e:
    # log error message from `e` or do something else
    return False
```
This way, if expiration is the only concern, it’s clear, and any other JWT-related issues can be handled as well. My guess is it's still possible that TokenInvalid can be raised here if non-JWT decodable thing is provided. Let me know what you think!

teaspoons-cli/teaspoons/cli.py
```diff
@@ -1,14 +1,30 @@
...
-import typer
-from commands.auth import auth_app
-from commands.pipelines import pipelines_app
+import click
 
-cli = typer.Typer()
-cli.add_typer(auth_app, name="auth", help="Authentication commands")
-cli.add_typer(pipelines_app, name="pipelines", help="Pipeline commands")
+from teaspoons import __version__
+from teaspoons.commands.auth_commands import auth as auth
```
fedorgrab, line 6:
```suggestion
from teaspoons.commands.auth_commands import auth
```

teaspoons-cli/teaspoons/cli.py
```diff
@@ -1,14 +1,30 @@
...
+from teaspoons.commands.pipelines_commands import pipelines as pipelines
```
fedorgrab, line 7:
```suggestion
from teaspoons.commands.pipelines_commands import pipelines
```

teaspoons-cli/teaspoons/commands/pipelines_commands.py
```diff
@@ -0,0 +1,30 @@
...
+@pipelines.command()
+def list():
+    """List all available pipelines"""
+    pipelines_list = pipelines_logic.list_pipelines()
+    click.echo(f"Found {len(pipelines_list)} available pipeline{'' if len(pipelines_list) == 1 else 's'}:")
+    for pipeline in pipelines_list:
+        click.echo(f"\t{pipeline.pipeline_name} - {pipeline.description}")
+
+
+@handle_api_exceptions
```
fedorgrab, line 24: A small thing, but might be useful in debugging: decorators in python are applied from bottom to top, so may be it'd be better to decorate with `handle_api_exception` first and then with click-related stuff. Or even better to transfer `handle_api_exception` decoration to `pipelines_logic.list_pipelines()`, this would make this function only responsible for cli stuff.

teaspoons-cli/teaspoons/config.py
```diff
@@ -8,15 +8,36 @@
...
+    #         self.client_info = OAuth2ClientInfo.from_oidc_endpoint(
+    #             self.config["OAUTH_OPENID_CONFIGURATION_URI"],
+    #             client_id=self.config["OAUTH_CLIENT_ID"],
+    #             scopes=[f"openid+email+profile+{self.config['OAUTH_CLIENT_ID']}"]
+    #         )
+    #     else:
+    #         self.client_info = None
+
     def __init__(self):
         self.config = dotenv_values("../.env")
```
KevinCLydon, line 24: Depending on how this is likely to be used, you could maybe have a config file in the home directory, like `~/.teaspoons-config` or something. Although I'm not sure if that would make the most sense for your use case.

teaspoons-cli/teaspoons/config.py
```diff
@@ -8,15 +8,36 @@
 class CliConfig:
```
KevinCLydon, line 8: I think this is a perfectly good way to handle config

teaspoons-cli/tests/logic/test_pipelines_logic.py
```diff
@@ -0,0 +1,70 @@
+# tests/logic/test_pipelines_logic.py
+
+from unittest.mock import patch, Mock
```
KevinCLydon, line 3: You might like [mockito](https://pypi.org/project/mockito/) if you're unfamiliar with it.  I'm not sure if that would make some of the mocking stuff easier.

teaspoons-cli/teaspoons/commands/pipelines_commands.py
mmorgantaylor: that's a good thought, thanks. we did originally have `handle_api_exception` on the inner (logic) function, but we moved it to this one because it deals more with how to handle/return the error to the user, which does feel like it's "cli stuff," and we didn't want the logic functions to have any knowledge of click (which we'd use to return the error messages). do you feel strongly that the error handling belongs on the logic function?

teaspoons-cli/teaspoons/config.py
mmorgantaylor: thank you!

teaspoons-cli/teaspoons/config.py
mmorgantaylor: i do like calling the file `.teaspoons-config` rather than the generic `.env`... i don't really know how to think about where it should live. currently we're providing the file in the repo, but it's a future task to figure out whether that's the best way to do it or if it should be generated or pulled from somewhere else. given that users are ultimately going to pull this package from PyPi, do you have advice on how we should provide these values?

teaspoons-cli/tests/logic/test_pipelines_logic.py
mmorgantaylor: oh thanks, that's a much nicer interface!

teaspoons-cli/teaspoons/commands/pipelines_commands.py
fedorgrab: Honestly, I'm not entirely certain. I wouldn’t strongly insist on it either way. Maybe the best way to think about this is to ask: if you were to use the `list_pipelines` function outside of the CLI, would you want this handling logic included? Just something to consider!

Also, one more thing if I am not mistaken, your current `handle_api_exceptions` doesn't use click, it just prints to the user through python's `print`.

.github/workflows/python-cli-build-and-test.yml
```diff
@@ -0,0 +1,84 @@
+name: Build and test Python CLI
+
+on:
```
jsotobroad, line 3: hmm do we know if this workflow fails does that affect any of the other workflows?  I assume if this fails the tag-publish workflow would still do its thing?

teaspoons-cli/teaspoons/auth_helper.py
```diff
@@ -42,20 +52,25 @@ def _validate_token(token: str) -> bool:
         # This is just to ensure the token is not expired
         jwt.decode(token, options={"verify_signature": False, "verify_exp": True})
         return True
+    except jwt.ExpiredSignatureError:
+        LOGGER.info("Token expired")
```
jsotobroad, line 56: is this something we want to log to the user as opposed to just taking them to the browser and that whole thing so its just assumed their token is expired

teaspoons-cli/teaspoons/auth_helper.py
```diff
@@ -42,20 +52,25 @@ def _validate_token(token: str) -> bool:
...
     except Exception as e:
+        LOGGER.error(f"Error validating token: {e}")
         return False
 
 
 def _clear_local_token(token_file: str):
     try:
         os.remove(token_file)
     except FileNotFoundError:
+        LOGGER.debug("No local token found to clean up")
```
jsotobroad, line 67: what level of logging do we plan to have for the CLI when published, ERROR + INFO?

teaspoons-cli/teaspoons/auth_helper.py
```diff
@@ -70,5 +85,5 @@ def _load_local_token(token_file: str) -> t.Optional[str]:
 def _save_local_token(token_file: str, token: str):
     # Create the containing directory if it doesn't exist
     os.makedirs(os.path.dirname(token_file), exist_ok=True)
-    with open(token_file, 'w') as f:
+    with open(token_file, "w") as f:
```
jsotobroad, line 88: i assume these `"` changes are all the linter doing its thing?

teaspoons-cli/teaspoons/cli.py
```diff
@@ -1,14 +1,29 @@
...
-cli = typer.Typer()
-cli.add_typer(auth_app, name="auth", help="Authentication commands")
-cli.add_typer(pipelines_app, name="pipelines", help="Pipeline commands")
+from teaspoons import __version__
+from teaspoons.commands.auth_commands import auth
+from teaspoons.commands.pipelines_commands import pipelines
+
+
+# Context settings for commands, for overwriting some click defaults
+CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
```
jsotobroad, line 12: just curious, what does click default as the `help_option_names` ?  just `-h`?

teaspoons-cli/teaspoons/cli.py
```diff
@@ -1,14 +1,29 @@
...
+
+LOGGER = logging.getLogger(__name__)
```
jsotobroad, line 14: is this necessary if we arent logging in the file (or are we logging and i dont see it)?

teaspoons-cli/teaspoons/client.py
```diff
@@ -1,29 +1,36 @@
...
 
     def __enter__(self):
+        cli_config = CliConfig()  # initialize the config from environment variables
         token = _load_local_token(cli_config.token_file)
-        if not token:
-            raise ValueError('Please authenticate first')
-        else:
-            return _get_api_client(token)
+        if not (token and _validate_token(token)):
+            LOGGER.info("You must authenticate first.")
```
jsotobroad, line 29: this should probably something like "Logging you in" or something since we're doing it for them as opposed to asking them to run another command

teaspoons-cli/teaspoons/utils.py
```diff
@@ -23,9 +29,10 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
         except ApiException as e:
             formatted_message = f"API call failed with status code {e.status} ({e.reason}): {json.loads(e.body)['message']}"
-            click.echo(formatted_message, err=True)
+            LOGGER.info(formatted_message)
```
jsotobroad, line 32: should this be `ERROR` instead?

teaspoons-cli/teaspoons/utils.py
```diff
@@ -23,9 +29,10 @@ def wrapper(*args, **kwargs):
...
             exit(1)
-        except ValueError as e:
-            click.echo(str(e), err=True)
+        except Exception as e:
+            LOGGER.info(str(e))
```
jsotobroad, line 35: same here with `ERROR`

teaspoons-cli/tests/commands/test_pipelines_commands.py
```diff
@@ -0,0 +1,105 @@
...
+    ).thenReturn(test_pipeline)
+
+    with caplog.at_level(logging.DEBUG):
+        result = runner.invoke(
+            pipelines_commands.pipelines, ["get-info", "test_pipeline"]
+        )
+
+    assert result.exit_code == 0
+    verify(pipelines_commands.pipelines_logic).get_pipeline_info("test_pipeline")
+    assert "test_pipeline" in caplog.text
```
jsotobroad, line 63: is there any way we can test the full string in some of these tests to guard against the issue we were seeing earlier where the logger wasnt properly set up in this testing suite.

Or even just a test that logs something and makes sure its formatted correctly

teaspoons-cli/teaspoons/cli.py
```diff
@@ -1,14 +1,29 @@
...
+
+
+@click.group(name="teaspoons", context_settings=CONTEXT_SETTINGS)
+@click.version_option(__version__)
+def cli():
+    logging.basicConfig(level=logging.INFO, format='%(message)s')
```
jsotobroad, line 20: we should add some coloring for error level messages to be fancy :D (totally dont do it in this pr but worth thinking about for the future)

teaspoons-cli/tests/test_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+def test_config():
+    mock_client_info = mock()
+    when(config.OAuth2ClientInfo).from_oidc_endpoint(...).thenReturn(mock_client_info)
+
+    test_config = config.CliConfig("tests/.test.config")
+
+    assert test_config.config["TEASPOONS_API_URL"] == "not-real"
+    assert test_config.config["SERVER_PORT"] == "12345"
+    assert test_config.config["OAUTH_OPENID_CONFIGURATION_URI"] == "https://dontcare"
```
jsotobroad, line 15: 😂

jsotobroad (commented):
some ux style question along with lack of python skill questions :)

.github/workflows/python-cli-build-and-test.yml
mmorgantaylor: yeah this is independent of the other workflows

.github/workflows/python-cli-build-and-test.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Set up python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Build the thick CLI client and, by dependency, the thin client library
```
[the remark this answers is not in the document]
mmorgantaylor: yes, assuming poetry

teaspoons-cli/pyproject.toml
```diff
@@ -4,17 +4,65 @@ version = "0.0.1"
...
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+    "generated"
+]
+# Same as Black.
+line-length = 88
```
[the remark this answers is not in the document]
mmorgantaylor: i assume they're the default - these were all suggestions from the ruff documentation

teaspoons-cli/teaspoons/auth_helper.py
mmorgantaylor: fair, we can make this a debug log

teaspoons-cli/teaspoons/auth_helper.py
mmorgantaylor: yeah INFO and up is the current setting. a DEBUG log will be picked up in tests. we don't yet have a way to run the cli directly with debug logging, can add that easily though

teaspoons-cli/teaspoons/cli.py
mmorgantaylor: looks like only `--help` - without this:

```
 ✗ ./teaspoons -h
Usage: teaspoons [OPTIONS] COMMAND [ARGS]...
Try 'teaspoons --help' for help.

Error: No such option: -h
```

teaspoons-cli/teaspoons/utils.py
mmorgantaylor: ah yes good catch, thanks!

teaspoons-cli/tests/commands/test_pipelines_commands.py
mmorgantaylor: i don't really care about the test logger being formatted correctly. i think if we want to test that the logs are formatted correctly we should add a test of `cli.py` (where the log formatting is set)

PR #149, opened 2024-10-30
TSPS-354 add pipeline quotas table

jsotobroad:
### Description 

Starting off the quotas work, adding the pipeline_quotas table which will be used in future PRs flesh out the quota work

Not really bothering with tests here since there is no logic yet to really test

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-354


mmorgantaylor (approved):
looks great, thanks!

PR #154, opened 2024-11-05
TSPS-367 fix: choose checkout-ref and tags based on github event type

mmorgantaylor:
### Description 

This is a fix to resolve the error we ran into [here](https://github.com/DataBiosphere/terra-scientific-pipelines-service/actions/runs/11690111192/job/32554380033?pr=153), where a run of Bump, Tag, Publish, Deploy on a PR failed because it tried to access a checkout-ref of a tag that was not actually created.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-367

### DIFF
--- a/.github/workflows/release-python-client.yml
+++ b/.github/workflows/release-python-client.yml
@@ -4,35 +4,13 @@ on:
   pull_request:
   workflow_call:
     inputs:
-      new-tag:
+      python-version:
+        required: true
+        type: string
+      checkout-ref:
         required: true
         type: string
 jobs:
-  params-gen:
-    runs-on: ubuntu-latest
-    outputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-python-client.yml
```diff
@@ -4,35 +4,13 @@ on:
   pull_request:
   workflow_call:
     inputs:
-      new-tag:
+      python-version:
```
mmorgantaylor, line 7: is it confusing to call this `python-version` even though it's coming from the new tag? it's being used as the python version but its source is the new tag, so now I'm rethinking renaming it.

mmorgantaylor: could rename it `new-tag-for-python-version`?

PR #162, opened 2024-11-18
TSPS-357 Add admin endpoint to get a user's quota and update their quota

jsotobroad:
### Description 

At some point we may want to increase a users quota above the default.  This admin endpoint makes it possible for us to do so if we want to.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-357

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -4,7 +4,7 @@ info:
   version: 1.0.0
 paths:
   # Admin APIs
-  /api/admin/v1/pipeline/{pipelineName}:
+  /api/admin/v1/pipelines/{pipelineName}:
     parameters:
       - $ref: '#/components/parameters/PipelineName'
     get:
@@ -48,14 +48,14 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
-  /api/admin/v1/quota/{pipelineName}/{userId}:
+  /api/admin/v1/quotas/{pipelineName}/{userId}:
     parameters:
       - $ref: '#/components/parameters/PipelineName'
       - $ref: '#/components/parameters/UserId'
     get:
-      summary: Get description for a given pipeline.
+      summary: Get quota for a given pipeline and user.
       tags: [ admin ]
-      operationId: getQuota
+      operationId: getQuotaForPipelineAndUser
       responses:
         200:
           description: Success
@@ -70,9 +70,9 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
     patch:
-      summary: Update attributes for a given pipeline.
+      summary: Update quota limit for a given pipeline and user.
       tags: [ admin ]
-      operationId: updateQuotaLimit
+      operationId: updateQuotaLimitForPipelineAndUser
       requestBody:
         required: true
         content:
@@ -403,7 +403,7 @@ components:
     UserId:
       name: userId
       in: path
-      description: A string identifier to used to identify a terra user
+      description: A string identifier to used to identify a Terra user
       required: true
       schema:
         type: string
@@ -929,7 +929,7 @@ components:
 
     UpdateQuotaLimitRequestBody:
       description: |
-        json object containing the admin provided information to update a users quota limit with
+        json object containing the admin provided information used to update a user's quota limit
       type: object
       required: [ quotaLimit ]
       properties:
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -48,7 +48,7 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
-  /api/admin/v1/quotas/{pipelineName}/{userId}:
+  /api/admin/v1/quota/{pipelineName}/{userId}:
     parameters:
       - $ref: '#/components/parameters/PipelineName'
       - $ref: '#/components/parameters/UserId'
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -48,6 +48,51 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
+  /api/admin/v1/quotas/{pipelineName}/{userId}:
+    parameters:
+      - $ref: '#/components/parameters/PipelineName'
+      - $ref: '#/components/parameters/UserId'
+    get:
+      summary: Get description for a given pipeline.
+      tags: [ admin ]
+      operationId: getQuota
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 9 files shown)


service/src/main/java/bio/terra/pipelines/service/QuotasService.java
```diff
@@ -75,4 +75,32 @@ public UserQuota updateQuotaConsumed(UserQuota userQuota, int newQuotaConsumed)
...
+
+  /**
+   * This method updates the quota limit for a given user quota. This should only be called from the
+   * Admin Controller
+   *
+   * @param userQuota - the user quota to update
+   * @param newQuotaLimit - the new quota limit
+   * @return - the updated user quota
+   */
+  public UserQuota updateQuotaLimit(UserQuota userQuota, int newQuotaLimit) {
```
mmorgantaylor, line 87: for clarity could we call this `adminUpdateUserQuotaLimit` ? thank you for mentioning that this is meant to be called from the admin controller.

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
```diff
@@ -75,4 +75,32 @@ public UserQuota updateQuotaConsumed(UserQuota userQuota, int newQuotaConsumed)
...
+    userQuota.setQuota(newQuotaLimit);
+    return userQuotasRepository.save(userQuota);
+  }
+
+  /**
+   * @param userId - the user id
+   * @param pipelineName - the pipeline name
+   * @return - true if the user quota is present, false otherwise
+   */
+  public boolean isUserQuotaPresent(String userId, PipelinesEnum pipelineName) {
```
mmorgantaylor, line 103: since this doesn't do any other authz checking, can we mark this as also must be called by an admin controller? (just in documentation / function name, as above)

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -74,6 +91,28 @@ public ResponseEntity<ApiAdminPipeline> updatePipeline(
...
+  public ResponseEntity<ApiAdminQuota> updateQuotaLimit(
+      String pipelineName, String userId, ApiUpdateQuotaLimitRequestBody body) {
+    final SamUser authedUser = getAuthenticatedInfo();
+    samService.checkAdminAuthz(authedUser);
+    PipelinesEnum validatedPipelineName =
+        PipelineApiUtils.validatePipelineName(pipelineName, logger);
+
+    // Check if a row exists for this user + pipeline. Throw an error if it doesn't
+    // A row should exist if a user has run into a quota issue before.
+    if (!quotasService.isUserQuotaPresent(userId, validatedPipelineName)) {
```
mmorgantaylor, line 104: curious about the thought process behind this approach, versus adding an option to `getQuotaForUserAndPipeline` that specifies whether to create a new row if the user isn't found and just reusing that method below?

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
```diff
@@ -75,4 +75,32 @@ public UserQuota updateQuotaConsumed(UserQuota userQuota, int newQuotaConsumed)
...
+  /**
+   * This method updates the quota limit for a given user quota. This should only be called from the
+   * Admin Controller
+   *
+   * @param userQuota - the user quota to update
+   * @param newQuotaLimit - the new quota limit
+   * @return - the updated user quota
+   */
+  public UserQuota updateQuotaLimit(UserQuota userQuota, int newQuotaLimit) {
+    if (newQuotaLimit < userQuota.getQuotaConsumed()) {
```
mmorgantaylor, line 88: so we would allow setting the newQuotaLimit to be equal to the user's current QuotaConsumed, thereby preventing any further activity, is that right? (sounds good to me)

mmorgantaylor (commented):
looking good, some questions and suggestions

common/openapi.yml
```diff
@@ -48,6 +48,51 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
+  /api/admin/v1/quota/{pipelineName}/{userId}:
```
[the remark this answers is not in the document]
jsotobroad: yea i was following the pattern we made for the admin pipeline path, should i update that to `pipelines` as well?

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
jsotobroad: hmmm well calling the other admin endpoint will actually create that row for the user cuz of how the`getQuotaForUserAndPipeline` function works so this is all kind of moot.  What scenario do we actually want to catch here?  Is it that the user has used our service before? is it that the user_id is an actual terra user_id?

we could write a get function that errors out if the row doesnt exist instead of generating a new one for these admin calls

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
jsotobroad: correct

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
mmorgantaylor: yeah I'm saying we could add an input to `getQuotaForUserAndPipeline` that's like `addRowIfAbsent=True` and toggle that functionality so we could reuse it, if that input is False and there's no row for that user then we throw. if that feels too complicated, that's fine, just an idea that I'm not at all tied to

common/openapi.yml
```diff
@@ -4,7 +4,7 @@ info:
   version: 1.0.0
 paths:
   # Admin APIs
-  /api/admin/v1/pipeline/{pipelineName}:
+  /api/admin/v1/pipelines/{pipelineName}:
```
mmorgantaylor, line 7: ah nice catch, thanks!

common/openapi.yml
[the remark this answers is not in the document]
mmorgantaylor: yeah thanks for fixing pipelines too!

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
mmorgantaylor: that's a good point, sounds good 👍

mmorgantaylor (approved):
love the updates!

PR #165, opened 2024-11-21
drop pipeline_name unique constraint from user_quotas table

jsotobroad:
### Description 

Fixing an issue that morgan found with the user_quotas table.  There is no reason for there to be a unique constraint on the pipeline name column of the user quotas table

added a test to make sure this is workign as intended

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
@@ -81,7 +81,7 @@ class QuotasServiceTest extends BaseEmbeddedDbTest {
             .findByUserIdAndPipelineName(TestUtils.TEST_USER_ID_1, PipelinesEnum.ARRAY_IMPUTATION)
             .isPresent());
 
-    // call service with second user and a new row should exist in user_quotas table
+    // call service with second user to test for unique constraints
     userQuota =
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
```diff
@@ -80,6 +80,11 @@ void addRowForNonExistentUserQuota() {
         userQuotasRepository
             .findByUserIdAndPipelineName(TestUtils.TEST_USER_ID_1, PipelinesEnum.ARRAY_IMPUTATION)
             .isPresent());
+
+    // call service with second user to test for unique constraints
+    userQuota =
+        quotasService.getOrCreateQuotaForUserAndPipeline(
+            TestUtils.TEST_USER_ID_2, PipelinesEnum.ARRAY_IMPUTATION);
```
mmorgantaylor, line 87: should we then assert that we can find that second user in the db? like add 
```
// assert second user + pipeline exists in the user_quotas table
    assertTrue(
        userQuotasRepository
            .findByUserIdAndPipelineName(TestUtils.TEST_USER_ID_2, PipelinesEnum.ARRAY_IMPUTATION)
            .isPresent());
```

mmorgantaylor (approved):
thanks for fixing this so fast! one suggestion for the test but 👍

PR #166, opened 2024-11-25
TSPS-350 address result_url issues and clean up cruft from teaspoons swap

jsotobroad:
### Description 
* the original purpose of this tickest was to correctly generate the result endpoint path for the get job and get pipeline run endpoints which this does
* after talkign abit we decided storing the result url in the pipeline_runs and the stairway "parameter" fields and returning does values does not give us the flexibility we may want in the future if we version either endpoint
* cleaned up some migration files that were no longer used in the swap to teaspoons and the subsequent squashing of our previous changesets.  I'm positive they're no longer needed but extra 👀 are always appreciated!

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-350


service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -262,21 +260,19 @@ private ApiAsyncPipelineRunResponse pipelineRunToApi(PipelineRun pipelineRun, Pi
...
    *
-   * @param ingressConfiguration configuration specifying the ingress domain name
+   * @param domainName the domain name from the ingress configuration
    * @param jobId identifier for the job
    * @return a string with the result endpoint URL
    */
-  public static String getAsyncResultEndpoint(
-      IngressConfiguration ingressConfiguration, HttpServletRequest request, UUID jobId) {
-    String endpointPath = "%s/result/%s".formatted(request.getServletPath(), jobId);
+  public static String getAsyncResultEndpoint(String domainName, UUID jobId) {
```
mmorgantaylor, line 271: especially since this can be static, I'd advocate for moving this to somewhere else (JobApiUtils would be fine by me but wherever you think makes sense) since it's used by multiple controllers, and it's really needed because we're dealing with async Stairway jobs that require us to provide the result path as part of the JobReport (i.e. it's not something inherent to PipelineRuns).

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
```diff
@@ -11,8 +11,8 @@ public class JobMapKeys {
   public static final String PIPELINE_ID = "pipeline_id";
   public static final String STATUS_CODE = "status_code";
   public static final String RESPONSE = "response"; // result or output of the job
-  public static final String RESULT_PATH =
-      "result_path"; // path to the result API endpoint for this job; only used for asynchronous
+  // domain name for the service, used to generate the URL for the result
+  public static final String DOMAIN_NAME = "domain_name";
```
mmorgantaylor, line 15: could this be an environment/configuration variable instead? is this ever going to change job to job within the same environment?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -117,7 +121,7 @@ public Map<String, Map<String, String>> preparePipelineRun(
   @WriteTransaction
   @SuppressWarnings("java:S1301") // allow switch statement with only one case
   public PipelineRun startPipelineRun(
-      Pipeline pipeline, UUID jobId, String userId, String description, String resultPath) {
+      Pipeline pipeline, UUID jobId, String userId, String description) {
```
mmorgantaylor, line 124: it's so much nicer not having to pass this value around!

service/src/test/java/bio/terra/pipelines/testutils/TestUtils.java
```diff
@@ -224,4 +223,8 @@ public static PipelineRun createNewPipelineRunWithJobIdAndUser(UUID jobId, Strin
         CONTROL_WORKSPACE_GOOGLE_PROJECT,
         CommonPipelineRunStatusEnum.PREPARING);
   }
+
+  public static String buildResultUrl(String jobId) {
```
mmorgantaylor, line 227: super nit, but for ease of reading tests and knowing this is a test method and not one used in the service, could we rename this to `buildTestResultUrl`?

mmorgantaylor (commented):
I don't mind this approach! it's a major improvement. just a couple suggestions/questions

service/src/main/java/bio/terra/pipelines/app/controller/JobApiUtils.java
```diff
@@ -80,14 +81,17 @@ public static ApiJobReport mapFlightStateToApiJobReport(FlightState flightState)
...
     return new ApiJobReport()
         .id(flightState.getFlightId())
         .description(description)
         .status(jobStatus)
         .statusCode(statusCode.value())
         .submitted(submittedDate)
         .completed(completedDate)
-        .resultURL(resultURL);
+        .resultURL(
+            PipelineRunsApiController.getAsyncResultEndpoint(
```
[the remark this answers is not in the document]
jsotobroad: yup that makes sense, can move it to JobApiUtils

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
jsotobroad: hmm i might be missing something, this is just the key name we're storing the domain value for a flight.  Why would we need this to be configurable?  the value being passed is coming from the ingress configuration if that's what you are referring to

service/src/main/java/bio/terra/pipelines/app/controller/JobApiUtils.java
```diff
@@ -80,14 +81,17 @@ public static ApiJobReport mapFlightStateToApiJobReport(FlightState flightState)
       }
     }
 
+    assert domainName != null;
```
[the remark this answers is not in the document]
jsotobroad: removed in favor of the flight validation function

mmorgantaylor (approved):
thanks!

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
jsotobroad: chatted with faces.  a better solution would be to grab the IngressConfiguration from a bean directly in the utils class we end up using this value in but that kind of adds other cruft to the code.  When we see more examples of us using flight params to pass information to the post flight functions then we should look into swapping over.  For one usage this is fine enough

PR #167, opened 2024-11-26
TSPS-381 Address PREPARING job response for result endpoint 

jsotobroad:
### Description 

I dont this endpoint should actually return a successful response for PREPARING jobs because nothing has been run yet.  I think this is the best way to handle that scenario by guiding the user down the correct path.

Screenshot 2024-11-26 at 1 05 29 AM

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-381

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
@@ -1,10 +1,12 @@
 package bio.terra.pipelines.app.controller;
 
+import bio.terra.common.exception.BadRequestException;
 import bio.terra.common.exception.NotFoundException;
 import bio.terra.common.iam.SamUser;
 import bio.terra.common.iam.SamUserFactory;
 import bio.terra.pipelines.app.configuration.external.IngressConfiguration;
 import bio.terra.pipelines.app.configuration.external.SamConfiguration;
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -160,6 +162,12 @@ public ResponseEntity<ApiAsyncPipelineRunResponse> getPipelineRunResult(
       throw new NotFoundException("Pipeline run %s not found".formatted(jobId));
     }
 
+    if (pipelineRun.getStatus().equals(CommonPipelineRunStatusEnum.PREPARING)) {
+      throw new BadRequestException(
```
mmorgantaylor, line 166: hmm, I disagree with this being a bad request / 400. it is an existing PipelineRun, it just has not been started yet. (calling the *job* endpoint should give a 400 since the *job* doesn't exist yet.) how about response code 201 ("[Created](https://restfulapi.net/http-status-codes/#2xx)") and keep the message you have here?

mmorgantaylor: discussed offline, the bad request response is ok here 👍

mmorgantaylor (approved):
sweet, as long as the CLI returns a reasonable response to the get details call (which I think it should) this looks good to me

PR #170, opened 2024-11-26
TSPS-382 remove pipelineName from serveral APIs that dont need it

jsotobroad:
### Description 

Note this does not change the result endpoint because that is being changed in https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/166 .  This PR also adds the pipeline name to the response of getAllPipelineRuns

Screenshot 2024-12-02 at 1 03 46 PM

Screenshot 2024-11-26 at 11 56 33 AM

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-382

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
@@ -22,6 +22,7 @@ import jakarta.servlet.http.HttpServletRequest;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -190,13 +192,19 @@ public ResponseEntity<ApiGetPipelineRunsResponse> getAllPipelineRuns(
     PageResponse<List<PipelineRun>> pageResults =
         pipelineRunsService.findPipelineRunsPaginated(maxLimit, pageToken, userId);
 
+    // convert list of pipelines to map of id to name
+    Map<Long, String> pipelineIdToNameMap =
```
mmorgantaylor, line 196: just so i understand, you do this here so that you only need to query the pipelines table once to get this mapping that you can use to populate the pipelineNames, rather than needing to query it for each response?

mmorgantaylor (approved):
looks great!

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
jsotobroad: exactly! a little overkill when we have one pipeline but should be futureproof

mmorgantaylor (approved):
looks great thanks!

PR #171, opened 2024-11-26
TSPS-370 Make version optional and defualt to max version for pipeline

jsotobroad:
### Description 

In this PR we 
* updated the prepare endpoint to optionally take in a pipelineVersion and if not supplied, automatically choose the latest version for the pieline
* updated the admin endpoints to required versions when getting and updating pipelines
* 

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-370


service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -47,8 +42,8 @@ public PipelinesApiController(
 
   private static final Logger logger = LoggerFactory.getLogger(PipelinesApiController.class);
 
-  private SamUser getAuthenticatedInfo() {
-    return samUserFactory.from(request, samConfiguration.baseUri());
+  private void getAuthenticatedInfo() {
```
mmorgantaylor, line 45: is this change because we don't actually use the `SamUser` output of this method? should we rename to something like `checkUserInSam` ?

service/src/main/java/bio/terra/pipelines/db/repositories/PipelinesRepository.java
```diff
@@ -15,4 +15,8 @@ public interface PipelinesRepository extends CrudRepository<Pipeline, Long> {
   Boolean existsByName(PipelinesEnum name);
 
   Pipeline findByName(PipelinesEnum name);
+
+  Pipeline findByNameAndVersion(PipelinesEnum name, Integer pipelineVersion);
+
+  Pipeline findFirstByNameOrderByVersionDesc(PipelinesEnum name);
```
mmorgantaylor, line 21: fancy!

service/src/test/java/bio/terra/pipelines/controller/PipelinesApiControllerTest.java
```diff
@@ -98,6 +104,55 @@ void getPipelineOk() throws Exception {
             .readValue(result.getResponse().getContentAsString(), ApiPipelineWithDetails.class);
 
     assertEquals(pipelineName, response.getPipelineName());
+    assertEquals(pipelineName, response.getPipelineName());
```
mmorgantaylor, line 107: is this a duplicate of the previous line?

mmorgantaylor (approved):
looks good! couple very minor things.

just to confirm, we currently don't have a way to add a new pipeline version (add a row to the pipelines table) via admin endpoint, right? via api, we can only update an existing pipeline/version combo?

service/src/test/java/bio/terra/pipelines/controller/PipelinesApiControllerTest.java
jsotobroad: 😬

jsotobroad:
> just to confirm, we currently don't have a way to add a new pipeline version (add a row to the pipelines table) via admin endpoint, right? via api, we can only update an existing pipeline/version combo?

correct it can not be done through the admin APIs.  I would strongly suggest adding a new pipeline/pipeline version should be a code change/migration.

PR #172, opened 2024-12-02
TSPS-375 reduce signed url durations

mmorgantaylor:
### Description 

Reduce the duration of signed urls:
- put / upload: reduced from 24 hours to 8 hours
- get / download: reduced from 24 hours to 1 hour

Note that the user must simply begin the process (upload or download) within the active duration of the signed url; the operation does not need to complete in that time.

Also: changing our artifactory repo to `libs-release` rather than `libs-snapshot-local`

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-375


jsotobroad (approved):

sarahgibs (approved):
Thank you, this looks great.

PR #173, opened 2024-12-03
TSPS-366 Add Content-Security-Policy header to Swagger page

mmorgantaylor:
### Description 

Add a Content-Security-Policy header to swagger.

Turns out that a lot in the `index.html` file gets overridden (e.g. the title there is `Terra Scientific Pipelines Service Swagger UI` but the swagger page title is just `Terra Scientific Pipelines Service`); adding the content header directly to that html doesn't work. Adding it in the Controller works.

Run locally:
image
The CSP also appears on the openapi.yml response headers.

Also: again trying to fix our deployment by updating the artifactory repo to `libs-release-local`.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-366

### DIFF
--- a/service/src/main/resources/templates/index.html
+++ b/service/src/main/resources/templates/index.html
@@ -1,107 +1,106 @@
 <!-- Copied from the swagger-ui static file -->
 <!DOCTYPE html>
 <html lang="en" xmlns:th="http://www.thymeleaf.org">
-  <head>
-    <meta charset="UTF-8">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'unsafe-inline'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; connect-src 'self'; form-action 'none';">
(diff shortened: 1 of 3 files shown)


jsotobroad (approved):
i dont know what fields that tag should have but im guessing thats good enough 🤷

service/src/main/resources/templates/index.html
```diff
@@ -1,106 +1,106 @@
 <!-- Copied from the swagger-ui static file -->
 <!DOCTYPE html>
 <html lang="en" xmlns:th="http://www.thymeleaf.org">
-  <head>
```
jsotobroad, line 4: are these changes purely formatting changes? (and waht did these changes?)

jsotobroad (approved):
just what question about formatting but looks good

service/src/main/resources/templates/index.html
mmorgantaylor: oh weird. yeah i think they're just whitespace changes? i didn't do them on purpose. i can revert

PR #176, opened 2024-12-04
TSPS-366 fix CSP header to work with b2c auth

mmorgantaylor:
### Description 

Update CSP header on swagger to not break b2c auth.

Also remove the redirect on the root swagger page.

### Jira Ticket
follow-on fix for https://broadworkbench.atlassian.net/browse/TSPS-366


jsotobroad (approved):
sweeet, thanks for the fix!!

PR #180, opened 2024-12-10
TSPS-388 Move description input from start endpoint to prepare endpoint

mmorgantaylor:
### Description 

The job description for each pipeline run is technically required for the flight, which is created on pipeline run start, but it was onerous and causing other problems to have to supply it then rather than with all the other inputs on pipeline run prepare, so we moved it to the prepare endpoint.

Note: Once this is merged, will need to update this CLI PR with the new thin client version: https://github.com/DataBiosphere/terra-scientific-pipelines-service-cli/pull/20

Calling prepare with a description:
image

Description is optional:
image

Start doesn't take a description:
image

Note that if no description is provided, it's still not included in the get all pipeline runs response and the CLI still doesn't like it, so [this PR in the cli repo](https://github.com/DataBiosphere/terra-scientific-pipelines-service-cli/pull/19) is still needed.
image

Companion PR to update the e2e test: https://github.com/broadinstitute/dsp-reusable-workflows/pull/60

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-388

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
@@ -62,7 +62,6 @@ class PipelineRunsServiceTest extends BaseEmbeddedDbTest {
   private final String testUserId = TestUtils.TEST_USER_ID_1;
   private final Long testPipelineId = TestUtils.TEST_PIPELINE_ID_1;
   private final String testWdlMethodVersion = TestUtils.TEST_WDL_METHOD_VERSION_1;
-  private final String testDescription = TestUtils.TEST_PIPELINE_DESCRIPTION_1;
   private final Map<String, Object> testPipelineInputs = TestUtils.TEST_PIPELINE_INPUTS;
   private final UUID testJobId = TestUtils.TEST_NEW_UUID;
   private final String testControlWorkspaceProject = TestUtils.CONTROL_WORKSPACE_BILLING_PROJECT;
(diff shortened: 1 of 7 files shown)


service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
```diff
@@ -296,7 +305,7 @@ void preparePipelineRunImputation() throws MalformedURLException {
 
     Map<String, Map<String, String>> formattedPipelineFileInputs =
         pipelineRunsService.preparePipelineRun(
-            testPipelineWithId, testJobId, testUserId, userPipelineInputs);
+            testPipelineWithId, testJobId, testUserId, userPipelineInputs, testDescription);
```
jsotobroad, line 308: probaby doesnt really matter but testDescription is pointing to the TestUtils.Pipeline_Description which is different from a PipelineRunDescription.  Not sure if we want to make a new static value in TestUtils for pipeline run description or use the pipeline one for both. It might make the tests a tiny bit more clearer and woudl catch if we accidentally were using the pipeline description where we should be using pipeline run description and vice versa

jsotobroad (approved):
One small comment on usage of the TestUtils pipeline description for both the pipeline and pipeline run description.  could merge w/o it i think though

service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
mmorgantaylor: ha! yeah I noticed this too and was too lazy to fix it (we were doing it elsewhere before this PR), but that's a good point. I'll fix, thanks!

PR #185, opened 2025-01-09
TSPS-340 return generic user facing message for 500 errors

jsotobroad:
### Description 

This will return a user facing generic support message for any 500 api response.  The actual error is still logged per usual

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-340

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
@@ -104,7 +104,7 @@ public class GlobalExceptionHandler {
   // -- catchall - log so we can understand what we have missed in the handlers above
   @ExceptionHandler(Exception.class)
   public ResponseEntity<ApiErrorReport> catchallHandler(Exception ex) {
-    logger.warn("Exception caught by catchall handler", ex);
+    logger.warn("Exception caught by catchall handler: {}", ex.getMessage());
     return buildApiErrorReport(ex, HttpStatus.INTERNAL_SERVER_ERROR, null, null);
   }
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -104,7 +104,7 @@ public ResponseEntity<ApiErrorReport> retryBackoffExceptionHandler(
   // -- catchall - log so we can understand what we have missed in the handlers above
   @ExceptionHandler(Exception.class)
   public ResponseEntity<ApiErrorReport> catchallHandler(Exception ex) {
-    logger.warn("Exception caught by catchall handler", ex);
+    logger.warn("Exception caught by catchall handler: {}", ex.getMessage());
```
jsotobroad, line 107: the `buildApiErrorReport` function already logs the full exception so i figured instead of doubling up in the logs we could instead just log the message here and connect that with the `buildApiErrorReport` log if necessary

mmorgantaylor:
isn't the example in the PR description not one we'd want to behave this way? that's where the user provided an unrecognized/invalid page token - that's a user error, shouldn't be a 500, right?

jsotobroad:
> isn't the example in the PR description not one we'd want to behave this way? that's where the user provided an unrecognized/invalid page token - that's a user error, shouldn't be a 500, right?

I think thats a different set of work if we want a better error for that issue specifically.  Like the ticket for that work should be "validate page_token values".  the issue is that its a runtime exception so its getting caught in our "catch all" exception handler which turns it into a 500 which then falls into the logic I added in this PR.

mmorgantaylor:
> > isn't the example in the PR description not one we'd want to behave this way? that's where the user provided an unrecognized/invalid page token - that's a user error, shouldn't be a 500, right?
> 
> I think thats a different set of work if we want a better error for that issue specifically. Like the ticket for that work should be "validate page_token values". the issue is that its a runtime exception so its getting caught in our "catch all" exception handler which turns it into a 500 which then falls into the logic I added in this PR.

yeah agreed, I was just thrown off by using that as an example. but it's probably not trivial to cause a 500 otherwise

PR #190, opened 2025-01-17
TSPS-424 add dsde qa slack notification to e2e test workflow

jsotobroad:
### Description 

We want #dsde-qa to be notified when our post deploy hook staging e2e workflow completes.  This is one way to do that

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-424

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -131,3 +131,11 @@ jobs:
       notify-slack-channels-upon-workflow-completion: "#terra-teaspoons-alerts"
     permissions:
       id-token: write
+
+  report-workflow-to-qa:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-e2e-tests.yaml
```diff
@@ -131,3 +131,11 @@ jobs:
       notify-slack-channels-upon-workflow-completion: "#terra-teaspoons-alerts"
     permissions:
       id-token: write
+
+  report-workflow-to-qa:
```
jsotobroad, line 135: not sure i like this option vs having a "slack channel" parameter that sends the notification to that channel only.  The only downside to that is that we wouldnt be able to look in one channel for all the e2e test results.

mmorgantaylor (approved):
looks good to me, thanks!

PR #191, opened 2025-01-21
TSPS-423 add smoke test  for teaspoons 

jsotobroad:
### Description 

For production readiness, we want to be able to run a smoke test against a deployment of teaspoons that will test the system quickly and with more confidence than a straight "/status" check.  This is a first pass at creating that test

This PR is related to https://github.com/broadinstitute/terra-github-workflows/pull/222

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-423

example manual run of the smoke_test.py script
Screenshot 2025-01-21 at 6 46 04 PM


smoke-test/tests/authenticated/teaspoons_pipeline_runs_list_tests.py
```diff
@@ -0,0 +1,11 @@
+from ..teaspoons_smoke_test_case import TeaspoonsSmokeTestCase
+
+
+class TeaspoonsPipelineRunsListTests(TeaspoonsSmokeTestCase):
```
mmorgantaylor, line 4: these are really quite clear, but it would make it even easier to quickly understand what we're testing if we add a docstring explaining what gets tested in each class of tests

mmorgantaylor (commented):
this looks super. what do you think of adding a query to the jobs endpoint (that'd make sure Stairway is working)? basically just make sure we get back a 200.

jsotobroad:
> this looks super. what do you think of adding a query to the jobs endpoint (that'd make sure Stairway is working)? basically just make sure we get back a 200.

i like it

mmorgantaylor (approved):
thank you, looks great!

PR #193, opened 2025-01-23
add some updates to readme

jsotobroad:
### Description 
Our e2e test failed during a staging release and DevOps would find it useful to have a list of services we depend on so they could better triage our test failures

### Jira Ticket
N/A


mmorgantaylor (approved):
thank you for all this!

PR #198, opened 2025-02-12
Add tracing, update TCL to 1.1.35 and Spring Boot to 3.4.2

mmorgantaylor:
### Description 

To support an eventual change we're making in TCL, we're separately updating the TCL version, since there are breaking changes to opentelemetry brought in with the updated TCL version.

Also taking this opportunity to enable tracing.

### Jira Ticket
No ticket for updating TCL
Add tracing ticket: https://broadworkbench.atlassian.net/browse/TSPS-107

### Checklist (provide links to changes)

- [ ] Updated external documentation (if applicable)
- [ ] Updated internal documentation (if applicable)
- [ ] Planned non patch version bump (if applicable)
- [ ] Updated CLI PR (if applicable)

PR #200, opened 2025-02-13
Fix tracing

mmorgantaylor:
### Description 

Upon enabling tracing (and updating to newer versions of opentelemetry and spring boot) in https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/198, we started seeing both errors in GCP logs saying `Failed to export spans. The request could not be executed. Full error message: Failed to connect to localhost/127.0.0.1:4318`, as well as traces being exported successfully to google. 

This PR resolves this (tested in a BEE; no errors in GCP logs but still do have tracing events) by [disabling otel tracing export](https://opentelemetry.io/docs/languages/java/configuration/#properties-exporters), since we're using google to do so instead.

Also renamed `pipelines` to `teaspoons` so that our service's traces are labeled `teaspoons`.

Also upped the status polling interval to 60 seconds from 30 seconds since we seemed to have excessive polling of Sam. Not sure if that'll really fix things but it was the original value we'd grabbed from elsewhere and was [decreased from 60 to 30](https://github.com/DataBiosphere/terra-scientific-pipelines-service/commit/c92ee286da93d39ffd6017e5937a8890efbf52e2) for some reason last year.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-107

- [x] Updated external documentation - README.md

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -4,7 +4,7 @@ env:
   db:
     host: jdbc:postgresql://${DATABASE_HOSTNAME:127.0.0.1}:5432
     init: ${INIT_DB:false}
-    teaspoons:
+    pipelines:
       name: ${DATABASE_NAME:pipelines_db}
       pass: ${DATABASE_USER_PASSWORD:dbpwd}
       user: ${DATABASE_USER:dbuser}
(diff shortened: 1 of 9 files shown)


jsotobroad (approved):
awesome thanks for all of these fixes.  What do we think about adding something to the readme (or one of the other documentation files) about tracing?  It doesnt look like other repos say anything about it so its prob not necessary.  Just popped into my head as a maybe.

mmorgantaylor:
@jsotobroad good call about the README - actually we already have a note about tracing there. I updated it slightly now that tracing is implemented - anything else you'd like me to add?

service/src/main/resources/application.yml
```diff
@@ -154,7 +154,7 @@ pipelines:
 
   status-check:
     enabled: true
-    pollingIntervalSeconds: 30
+    pollingIntervalSeconds: 60
```
gbggrant, line 157: a comment

PR #204, opened 2025-02-20
TSPS-384 rename wdl specific columns

jsotobroad:
### Description 
We want to update our tables / api to not be so wdl specific where possible so its easier to add non wdl pipelines in the future w/o having to change terminology as much

* rename wdl_method_version to tool_version and wdl_method_name to tool_name in pipelines and pipeline_runs tables
* also update deprecated MockBeans annotation to MockitoBean

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-384

PR to update e2e test - https://github.com/broadinstitute/dsp-reusable-workflows/pull/71
- [x] Updated CLI PR (if applicable)
https://github.com/DataBiosphere/terra-scientific-pipelines-service-cli/pull/43

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
+++ b/service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
@@ -44,11 +44,11 @@ public class Pipeline {
   @Column(name = "wdl_url")
   private String wdlUrl;
 
-  @Column(name = "wdl_method_name")
-  private String wdlMethodName;
+  @Column(name = "tool_name")
+  private String toolName;
 
-  @Column(name = "wdl_method_version")
-  private String wdlMethodVersion;
+  @Column(name = "tool_version")
+  private String toolVersion;
 
   @Column(name = "workspace_id")
   private UUID workspaceId;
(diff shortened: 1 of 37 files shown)


service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
```diff
@@ -81,8 +81,8 @@ public Pipeline(
       String description,
       String pipelineType,
       String wdlUrl,
```
mmorgantaylor, line 83: do we think it doesn't make sense to rename wdlUrl as well?

mmorgantaylor (commented):
looks great, one question about renaming the wdlUrl field

service/src/main/java/bio/terra/pipelines/db/entities/Pipeline.java
jsotobroad: i think that field is actually specific to a wdl pipeline so i think its ok to keep that as is.  Or we rename it if we feel like a url is necessary to set up the config for different types of pipelines but i cant think of what url like CAS would need for setup.  I think eventually we'll want a whole table for each type of pipeline and stuff liek wdlUrl would go there but no need to split that out now until we see a second pipeline type

mmorgantaylor: makes sense to me, thanks!

mmorgantaylor (approved):
thanks for making these changes!

PR #205, opened 2025-02-25
TSPS-405 Add quota units to pipeline quotas

mmorgantaylor:
### Description 

Each pipeline should have its quota unit defined. This PR adds a `quota_units` field to the `pipeline_quotas` table in the teaspoons database; the string value of `QuotaUnitsEnum` is meant to be the plural version of the unit, e.g. "samples" rather than "sample". We include this field in the `QuotaWithDetails` API response that users receive when they query their quota status for a pipeline.

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-405

- [ ] Updated CLI PR: https://github.com/DataBiosphere/terra-scientific-pipelines-service-cli/pull/45

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/QuotasController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/QuotasController.java
@@ -51,15 +51,17 @@ public class QuotasController implements QuotasApi {
     UserQuota userQuota =
         quotasService.getOrCreateQuotaForUserAndPipeline(
             user.getSubjectId(), validatedPipelineName);
+    String quotaUnits = quotasService.getPipelineQuota(validatedPipelineName).getQuotaUnits();
 
-    return new ResponseEntity<>(quotasToApiQuotaWithDetails(userQuota), HttpStatus.OK);
+    return new ResponseEntity<>(quotasToApiQuotaWithDetails(userQuota, quotaUnits), HttpStatus.OK);
   }
 
-  static ApiQuotaWithDetails quotasToApiQuotaWithDetails(UserQuota userQuota) {
+  static ApiQuotaWithDetails quotasToApiQuotaWithDetails(UserQuota userQuota, String quotaUnits) {
 
     return new ApiQuotaWithDetails()
         .pipelineName(userQuota.getPipelineName().getValue())
         .quotaConsumed(userQuota.getQuotaConsumed())
-        .quotaLimit(userQuota.getQuota());
+        .quotaLimit(userQuota.getQuota())
+        .quotaUnits(quotaUnits);
   }
 }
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/QuotasController.java
```diff
@@ -51,15 +51,17 @@ public ResponseEntity<ApiQuotaWithDetails> getQuotaForPipeline(String pipelineNa
     UserQuota userQuota =
         quotasService.getOrCreateQuotaForUserAndPipeline(
             user.getSubjectId(), validatedPipelineName);
+    String quotaUnits = quotasService.getPipelineQuota(validatedPipelineName).getQuotaUnits();
```
mmorgantaylor, line 54: could alternatively make a dedicated method just to retrieve the quota units from the db

jsotobroad (commented):
clean.  just had one comment about possibly using an enum to keep the column values as we expect

service/src/main/resources/db/changesets/20250225_add_quota_units.yaml
```diff
@@ -0,0 +1,28 @@
...
+                   type: text
+                   constraints:
+                     nullable: true
+         # update quota_units value for array_imputation pipeline
+         - update:
+             tableName: pipeline_quotas
+             columns:
+               - column:
+                  name: quota_units
+                  value: 'samples'
```
jsotobroad, line 23: should we update the e2e to test for this too?  prob not necessary

mmorgantaylor: we don't currently check all the fields returned from the quota endpoint... i could make the change to ask for all of them, but currently we only check the value of `response['quotaConsumed']`

service/src/main/java/bio/terra/pipelines/db/entities/converters/QuotaUnitsConverter.java
```diff
@@ -0,0 +1,19 @@
...
+
+import bio.terra.pipelines.common.utils.QuotaUnitsEnum;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+// inspired by https://www.baeldung.com/jpa-persisting-enums-in-jpa
+@Converter(autoApply = true)
+public class QuotaUnitsConverter implements AttributeConverter<QuotaUnitsEnum, String> {
+  @Override
+  public String convertToDatabaseColumn(QuotaUnitsEnum quotaUnitsEnum) {
```
mmorgantaylor, line 11: this is the one place that's not being tested. since we never actually use code to set this value in the db (we use liquibase), i'm ok with that, but i can construct a test if that would be better (we don't have explicit tests of these kinds of classes currently)

jsotobroad: yea i think thats more than fine to not test this.  We could prob even throw like a "not implemented error dont call me" here and it wouldnt be an issue.  that could be the solution :P

jsotobroad (approved):

PR #89, opened 2024-06-07
TSPS-257, TSPS-243 add Stairway retries to WDS and CBAS steps, also Don't alert Sentry for NoResourceFoundException

mmorgantaylor:
### Description 

- TSPS-257: add Stairway retry logic to steps that call CBAS or WDS
- TSPS-243: filter out NoResourceFoundExceptions from Sentry (still will be logged as errors, plus a warning that it wasn't sent to sentry)

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-257
https://broadworkbench.atlassian.net/browse/TSPS-243

PR #206, opened 2025-02-26
TSPS-345 Mute sentry alerts for HttpRequestMethodNotSupportedException

mmorgantaylor:
### Description 

We get Sentry alerts often for seemingly spam/bot calls to existing paths with invalid methods (POST, PROPFIND). Similar to https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/89, we don't send these to Sentry but do log that we did so.  

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-345

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
@@ -41,6 +41,7 @@ public class GlobalExceptionHandler {
   // -- validation exceptions - we don't control the exception raised
   @ExceptionHandler({
     MethodArgumentTypeMismatchException.class,
+    HttpRequestMethodNotSupportedException.class,
     IllegalArgumentException.class,
     NoHandlerFoundException.class
   })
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -41,7 +41,6 @@ public ResponseEntity<ApiErrorReport> errorReportHandler(ErrorReportException ex
   // -- validation exceptions - we don't control the exception raised
   @ExceptionHandler({
     MethodArgumentTypeMismatchException.class,
-    HttpRequestMethodNotSupportedException.class,
```
mmorgantaylor, line 44: this makes me a little nervous. we don't have any instances of this error in Sentry in dev in the last 30 days but this would be like a user doing something wrong, so not sure if this is going to mask something that the user is doing. but I think that it'll be ok - if that's the case, the user will still get the 4xx response and we can investigate logs if it's confusing to them

jsotobroad (commented):
😈

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
mmorgantaylor: my understanding is if we keep it, those errors will always be caught by this rather than later on and we'll still get sentry alerts for it. unless we move the sentry filtering thing into this method, maybe that's the answer?

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
```diff
@@ -115,9 +114,11 @@ private ResponseEntity<ApiErrorReport> buildApiErrorReport(
       @Nullable String messageForApiErrorReport) {
     // logging 4** & 5** errors to sentry
     if (statusCode.is5xxServerError() || statusCode.is4xxClientError()) {
-      if (ex.getClass() == NoResourceFoundException.class) {
+      if (ex instanceof NoResourceFoundException
+          || ex instanceof HttpRequestMethodNotSupportedException) {
```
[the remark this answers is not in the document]
mmorgantaylor: seems we should pick one place to ignore sentry errors, either here or in sentry, but not both.

service/src/main/java/bio/terra/pipelines/app/controller/GlobalExceptionHandler.java
mmorgantaylor: ah i misunderstood, you're totally right - will revert this

jsotobroad (approved):
chatted in person

PR #251, opened 2025-09-08
[TSPS-599] Use JPA for pipeline run history pagination and sorting

MatthewBemis:
### Description 

Note that this only moves over pipeline runs, not jobs.

Related draft PR: switch the UI over to v2 and take advantage of sorting: https://github.com/DataBiosphere/terra-ui/pull/5413

### DIFF
--- a/service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/PipelineRunsServiceTest.java
@@ -627,12 +627,12 @@ class PipelineRunsServiceTest extends BaseEmbeddedDbTest {
           org.mockito.Mockito.mock(PipelineRunsRepository.class);
       ArgumentCaptor<Pageable> pageableCaptor = ArgumentCaptor.forClass(Pageable.class);
 
-      PipelineRunsService pipelineRunsService =
+      PipelineRunsService mockPipelineRunsService =
           new PipelineRunsService(
               mockJobService, pipelineInputsOutputsService, mockPipelineRunsRepository, null, null);
 
       // query with null sort params, should default to created DESC
-      pipelineRunsService.findPipelineRunsPaginated(0, 10, "created", null, testUserId);
+      mockPipelineRunsService.findPipelineRunsPaginated(0, 10, "created", null, testUserId);
 
       verify(mockPipelineRunsRepository).findAllByUserId(eq(testUserId), pageableCaptor.capture());
 
@@ -649,12 +649,12 @@ class PipelineRunsServiceTest extends BaseEmbeddedDbTest {
           org.mockito.Mockito.mock(PipelineRunsRepository.class);
       ArgumentCaptor<Pageable> pageableCaptor = ArgumentCaptor.forClass(Pageable.class);
 
-      PipelineRunsService pipelineRunsService =
+      PipelineRunsService mockPipelineRunsService =
           new PipelineRunsService(
               mockJobService, pipelineInputsOutputsService, mockPipelineRunsRepository, null, null);
 
       // query with null sort property, should default to created
-      pipelineRunsService.findPipelineRunsPaginated(0, 10, null, "DESC", testUserId);
+      mockPipelineRunsService.findPipelineRunsPaginated(0, 10, null, "DESC", testUserId);
 
       verify(mockPipelineRunsRepository).findAllByUserId(eq(testUserId), pageableCaptor.capture());
 


service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -361,4 +387,40 @@ public PageResponse<List<PipelineRun>> findPipelineRunsPaginated(
...
+    }
+    try {
+      return Sort.Direction.valueOf(sortDirection.toUpperCase());
+    } catch (IllegalArgumentException e) {
+      throw new BadRequestException(
+          "Invalid sort direction: %s. Valid values are ASC or DESC".formatted(sortDirection));
+    }
+  }
+
+  private String validateSortProperty(String sortProperty) {
```
MatthewBemis, line 403: there's a bit of a quirk here that requires the caller of this to know the internal representation of a PipelineRun. in `getAllPipelineRuns` (v1 and v2), we return a ApiPipelineRun which changes some of the property names:

i.e. created -> timeSubmitted

in order to sort on the field, the caller needs to specify what's stored in the `PipelineRun` table, `created` in this case. but in the response they get, they'll see `timeSubmitted`. so it might be confusing.

i'd love to get thoughts on this. i could add a step to translate the user-facing fields and turn them into a field that actually exists on PipelineRun? It feels a bit messy though...

service/src/test/java/bio/terra/pipelines/controller/PipelineRunsApiControllerTest.java
```diff
@@ -7,9 +7,7 @@
 import static org.junit.jupiter.api.Assertions.assertNull;
```
MatthewBemis, line 7: at this point, you probably want to start reviewing with whitespace changes disabled. there were some nesting changes which i'll explain below...

https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/251/files?w=1

service/src/test/java/bio/terra/pipelines/controller/PipelineRunsApiControllerTest.java
```diff
@@ -721,176 +723,249 @@ void getPipelineRunResultNotFound() throws Exception {
...
-        new ObjectMapper()
-            .readValue(result.getResponse().getContentAsString(), ApiGetPipelineRunsResponse.class);
-
-    assertNull(response.getPageToken());
-    assertEquals(4, response.getResults().size());
-    assertEquals(4, response.getTotalResults());
+  @Nested
+  @Deprecated
+  @DisplayName("getAllPipelineRuns v1 tests")
+  class GetAllPipelineRunsV1Tests {
```
MatthewBemis, line 791: what do y'all think of this way of organizing tests for deprecated endpoints?

i think nesting them all in a subclass makes it easier to know which tests we'll need to rip out once the CLI moves over to v2 and we're ready to remove v1 entirely. but i know there are a million different ways to accomplish this so i'm curious if anyone has any different ideas

mmorgantaylor: this makes a lot of sense to me!

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -321,8 +324,30 @@ public PipelineRun markPipelineRunSuccessAndWriteOutputs(
    */
   public void markPipelineRunFailed(UUID jobId, String userId) {
     PipelineRun pipelineRun = getPipelineRun(jobId, userId);
-    pipelineRun.setStatus(CommonPipelineRunStatusEnum.FAILED);
     pipelineRunsRepository.save(pipelineRun);
+    pipelineRun.setStatus(CommonPipelineRunStatusEnum.FAILED);
```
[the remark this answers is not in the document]
mmorgantaylor: this is probably why `StairwaySetPipelineRunStatusHookTest` is failing

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -250,6 +252,56 @@ public ResponseEntity<ApiGetPipelineRunsResponse> getAllPipelineRuns(
     return new ResponseEntity<>(apiGetPipelineRunsResponse, HttpStatus.OK);
   }
 
+  // TODO: check indices
```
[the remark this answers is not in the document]
jsotobroad: yea we should chat about this in parking lot

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
jsotobroad: yea im thinking either have the UI handle this or maybe we just keep the returned values the same as the able column names?  Like just return "created" to the user".  Lots of cool parking lots to talk about!

MatthewBemis: it's easy to brush over in the ui since the sort field is manually specified per column.

here's where it'll specify `created` instead of `timeSubmitted` https://github.com/DataBiosphere/terra-ui/pull/5413/files#diff-3affab90042cf5362b6a8608bb35fcf2ed8e25e942086e657a05ee8db7fb1080R188

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -334,6 +359,7 @@ public void markPipelineRunFailed(UUID jobId, String userId) {
    * @return - a PageResponse containing the list of records in the current page and the page tokens
    *     for the next and previous page if applicable
    */
+  @Deprecated
```
jsotobroad, line 362: can we create a ticket to remove this endpoint and all relevant code (including the pagination codue this method calls.  Thanks for these `@Deprecated` annotations!

jsotobroad (commented):
this is uhhh, way nicer than what we had, cant wait to get rid of v1 lol

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -361,4 +387,33 @@ public PageResponse<List<PipelineRun>> findPipelineRunsPaginated(
...
+    }
+  }
+
+  private String validateSortProperty(String sortProperty) {
+    // default to sorting by created
+    if (sortProperty == null) {
+      return "created";
+    }
+
+    List<String> allowedSortProperties = List.of("created", "updated", "quotaConsumed");
```
jsotobroad, line 409: is there any way to verify that these fields exist in the underlying entity being paginated (PipelineRun)?  Might be nice to just catch ourselves if we change the db tables but forget to update this allowed list.  Could be here or possibly a test.

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
MatthewBemis: I've just filed [TSPS-622](https://broadworkbench.atlassian.net/browse/TSPS-624), [623](https://broadworkbench.atlassian.net/browse/TSPS-623), and [624](https://broadworkbench.atlassian.net/browse/TSPS-624) which relate to this

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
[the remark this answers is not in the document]
MatthewBemis: discussed in post: we're only going to allow sorting on `submitted`, `created`, and `quotaConsumed`

(that way we don't have to add indexes to columns like `description`)

and for things like jobId and status, those will be better suited for a filter/search, rather than a sort.

PR #260, opened 2025-10-01
[TSPS-611] Add input validation for strings and file names

MatthewBemis:
### Description 

Adds input validation for strings, string arrays, files, and file arrays.

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-611

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
+++ b/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
@@ -226,7 +226,9 @@ public enum PipelineVariableTypesEnum {
   public abstract String validate(PipelineInputDefinition pipelineInputDefinition, Object value);
 
   private static final String NOT_NULL_OR_EMPTY_ERROR_MESSAGE = "%s must not be null or empty";
-  private static final Pattern VALID_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9_.=\\\\\\-/]+$");
+  // this regex only allows alphanumeric characters, dashes, underscores, periods, equal signs, and
+  // forward and backward slashes
+  private static final Pattern VALID_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9_.=\\\\/-]+$");
 
   @SuppressWarnings(
       "java:S1168") // Disable "Empty arrays and collections should be returned instead of null"
--- a/service/src/test/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnumTest.java
+++ b/service/src/test/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnumTest.java
@@ -104,6 +104,7 @@ class PipelineVariableTypesEnumTest extends BaseTest {
 
         // STRING
         arguments(stringInputDefinition, "IAmAString", "IAmAString", null),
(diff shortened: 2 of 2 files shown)


service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -206,6 +226,7 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
   public abstract String validate(PipelineInputDefinition pipelineInputDefinition, Object value);
 
   private static final String NOT_NULL_OR_EMPTY_ERROR_MESSAGE = "%s must not be null or empty";
+  private static final Pattern VALID_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9_.=\\-/]+$");
```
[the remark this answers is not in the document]
jsotobroad: i believe this regex is also allowing the `\` character (or my regex skills are rusty).  if so we should update the error message

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -92,6 +98,10 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
       if (stringCastValue == null || !(stringCastValue.endsWith(fileSuffix))) {
         return "%s must be a path to a file ending in %s".formatted(fieldName, fileSuffix);
       }
+      if (!VALID_STRING_PATTERN.matcher(stringCastValue).matches()) {
+        return "%s must be a file path containing only alphanumeric characters or the following symbols: - _ . = /"
```
jsotobroad, line 102: i think you can use the same error string here as in the string error message (in this case it's not really relevant if its a file path) just to help with refactoring some of these strings out

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -23,9 +24,14 @@ public <T> T cast(String fieldName, Object value, TypeReference<T> typeReference
     @Override
     public String validate(PipelineInputDefinition pipelineInputDefinition, Object value) {
       String fieldName = pipelineInputDefinition.getName();
-      if (cast(fieldName, value, new TypeReference<String>() {}) == null) {
+      String castValue = cast(fieldName, value, new TypeReference<String>() {});
+      if (castValue == null) {
         return "%s must be a string".formatted(fieldName);
       }
+      if (!VALID_STRING_PATTERN.matcher(castValue).matches()) {
```
jsotobroad, line 31: i know your ticket talked about trimming before validating for spaces.  Does trimming already happen before this validation or is that something we should do here?

MatthewBemis: `FILE` reuses the `cast` method from `STRING` (see [here](https://github.com/DataBiosphere/terra-scientific-pipelines-service/blob/main/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java#L110-L126)), so it gets the same trimming done. there are also a few cases for it in both the `FILE` and `FILE_ARRAY` cast+validation tests

PR #261, opened 2025-10-02
TSPS-641 modify data table entity name to inclue pieline version

jsotobroad:
### Description 
Doing some pre work for when we have multiple versions per pipeline.  This PR add the version to the data table entity name and the related changes for that to work.

I tested running workflows with the old flight, killing the server, updating the code, then running the server.  If the old flight was passt the modified steps, it succeeded.  If the old flight was stopped beore the modified steps, then it failed when it got to those steps

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-641

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
+++ b/service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
@@ -1,6 +1,7 @@
 package bio.terra.pipelines.dependencies.rawls;
 
 import bio.terra.common.exception.InternalServerErrorException;
+import bio.terra.pipelines.common.utils.PipelinesEnum;
 import bio.terra.pipelines.db.entities.PipelineInputDefinition;
 import bio.terra.pipelines.db.entities.PipelineOutputDefinition;
 import bio.terra.pipelines.dependencies.common.HealthCheck;
@@ -313,5 +314,10 @@ public class RawlsService implements HealthCheck {
             throw new RawlsServiceApiException(e);
           }
         });
+  }
+
+  public static String createDataTableEntityName(
+      PipelinesEnum pipelineName, Integer pipelineVersion) {
+    return pipelineName.getValue() + "_v" + pipelineVersion;
   }
 }
--- a/service/src/main/java/bio/terra/pipelines/stairway/flights/imputation/v20251002/RunImputationGcpJobFlight.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/flights/imputation/v20251002/RunImputationGcpJobFlight.java
@@ -1,0 +1,160 @@
+package bio.terra.pipelines.stairway.flights.imputation.v20251002;
+
+import bio.terra.pipelines.app.common.MetricsUtils;
+import bio.terra.pipelines.common.utils.FlightBeanBag;
+import bio.terra.pipelines.common.utils.FlightUtils;
+import bio.terra.pipelines.common.utils.PipelinesEnum;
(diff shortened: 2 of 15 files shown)


service/src/main/java/bio/terra/pipelines/stairway/flights/imputation/v20251002/RunImputationGcpJobFlight.java
```diff
@@ -0,0 +1,160 @@
...
+
+  public RunImputationGcpJobFlight(FlightMap inputParameters, Object beanBag) {
+    super(inputParameters, beanBag);
+    final FlightBeanBag flightBeanBag = FlightBeanBag.getFromObject(beanBag);
+
+    FlightUtils.validateRequiredEntries(
+        inputParameters,
+        JobMapKeys.USER_ID,
+        JobMapKeys.PIPELINE_NAME,
+        JobMapKeys.PIPELINE_VERSION,
```
jsotobroad, line 44: this was the "big" change

service/build.gradle
```diff
@@ -84,6 +84,9 @@ dependencies {
     // sentry
     implementation 'io.sentry:sentry:7.3.0'
 
+    // jersey hk2 for dependency injection
+    implementation 'org.glassfish.jersey.inject:jersey-hk2'
```
jsotobroad, line 88: was seeing a lot of 
```
Jersey-HK2 module is missing. Falling back to injection-less client. Injection may not be supported on the client.
```
warn logs in our logs that this should get rid of for seemingly no downside?

mmorgantaylor: thank you for addressing this!

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -314,4 +315,9 @@ static <T> T executionWithRetryTemplate(RetryTemplate retryTemplate, RawlsAction
           }
         });
   }
+
+  public static String createDataTableEntityName(
```
mmorgantaylor, line 319: take or leave: in case we want to change this at all in future, what do you think of making the input just a Pipeline object and extracting the name and version out of it inside the method?
edit: after seeing the rest of the PR, i think the answer is that we have pipelineName and (now) pipelineVersion in the working map (input params), and not the Pipeline object. that makes sense. we could consider a refactor to just have the Pipeline object in the working map but doesn't make sense to do it here.
all to say, looks great!

service/src/main/java/bio/terra/pipelines/stairway/steps/common/SubmitCromwellSubmissionStep.java
```diff
@@ -95,7 +100,7 @@ public StepResult doStep(FlightContext flightContext) {
     SubmissionRequest submissionRequest =
         new SubmissionRequest()
             .entityName(flightContext.getFlightId())
-            .entityType(pipelineName.getValue())
+            .entityType(dataTableEntityName)
```
mmorgantaylor, line 103: this is so much clearer now

mmorgantaylor (approved):
looks great!

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
jsotobroad: yea i had the same thought process you put into text here 😂 . I do think refactoring will be worth doing if we continue adding pipeline fields to flights

MatthewBemis (approved):

PR #366, opened 2026-04-23
TSPS-873 add deserializer function so internal step exception

jsotobroad:
### Description 
we had a bug in our code where internal step exceptions would be "double" wrapped when being returned to the user.  So instead of 
```
An error occurred while running the job. Not all workflows succeeded in submission. Please contact support for help.
```
it was returning
```
An error occurred while running the job. An error occurred while running the job. Not all workflows succeeded in submission. Please contact support for help. Please contact support for help.
```
In the stairway database it was the correct string but when being deserialized it was being wrapped.  The bots figured out the rest and which method we needed to add so the stairway exception serializer would do the appropriate thing instead of falling back to the string constructor which would wrap the string again on the way out.

before
Screenshot 2026-04-23 at 1 34 48 PM

after
Screenshot 2026-04-23 at 1 33 19 PM

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-873

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/stairway/steps/exception/InternalStepException.java
+++ b/service/src/main/java/bio/terra/pipelines/stairway/steps/exception/InternalStepException.java
@@ -1,6 +1,7 @@
 package bio.terra.pipelines.stairway.steps.exception;
 
 import bio.terra.common.exception.InternalServerErrorException;
+import java.util.List;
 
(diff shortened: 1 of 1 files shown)


mmorgantaylor (approved):
nice, thank you for fixing this!

MatthewBemis (approved):

salonishah11 (approved):
Nice!🥇

PR #371, opened 2026-04-29
TSPS-753 Remove deprecated v1 result endpoint

mmorgantaylor:
### Description 

In https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/288 we added a v2 /result endpoint and deprecated the v1 /result endpoint. Here we remove the code associated with the v1 endpoint.

Logs show that the last time the v1 /result endpoint was called in production was February 4, 2026. We are safe to remove it.

This will be a MAJOR version bump.

Related PR to update e2e test: https://github.com/broadinstitute/terra-github-workflows/pull/254

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-753


common/openapi.yml
```diff
@@ -204,13 +204,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/AsyncPipelineRunResponse'
+                $ref: '#/components/schemas/AsyncPipelineRunResponseV2'
```
jsotobroad, line 207: im guessing this would be a breaking change but people _shouldnt_ be callign this endpoint after its been started

common/openapi.yml
```diff
@@ -204,13 +204,13 @@ paths:
...
         202:
           description: PipelineRun is running
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/AsyncPipelineRunResponse'
+                $ref: '#/components/schemas/AsyncPipelineRunResponseV2'
```
jsotobroad, line 213: correct me if im wrong but the parts that get returned by start for `AsyncPipelineRunResponse` and `AsyncPipelineRunResponseV2` are the same right?

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -742,7 +638,7 @@ private JobApiUtils.AsyncJobResult<String> getRunningOrFailedJobResult(PipelineR
                   .completed(pipelineRun.getUpdated().toString())
                   .resultURL(
                       getAsyncResultEndpoint(
-                          ingressConfiguration.getDomainName(), jobId, 1))) // 1 is resultApiVersion
+                          ingressConfiguration.getDomainName(), jobId, 2))) // 2 is resultApiVersion
```
jsotobroad: im guessing you updated all the places we used to grab version 1, thanks!

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
```diff
@@ -890,7 +890,8 @@ && getFileLocationType(rawOrCustomValue) == FileLocationTypeEnum.LOCAL) {
    *
    * @param jobId UUID
    * @param allInputDefinitions List<PipelineInputDefinition>
-   * @param userProvidedPipelineInputs Map<String, Object>
+   * @param userProvidedPipelineInputs Map<String, Object> already populated with default values for
```
jsotobroad: whats this change for?

service/src/test/java/bio/terra/pipelines/controller/PipelineRunsApiControllerTest.java
```diff
@@ -1268,405 +1268,6 @@ void getPipelineRunResultNotFound() throws Exception {
     }
   }
 
-  @Nested
```
jsotobroad, line 1271: this feels great 😂

jsotobroad (approved):
sweeeeet

common/openapi.yml
mmorgantaylor: that's right, V2 changed what's in the outputs section (output file names, not signed urls), which isn't included when the job's just been started / is still running.

common/openapi.yml
mmorgantaylor: i don't follow, can you say more?

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
mmorgantaylor: yeah I hope so. I was thinking of making a constant for this so we don't have to find all these places when we want to update it

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
mmorgantaylor: oh this was a change from a different PR that was accidentally committed here. will remove, thanks

common/openapi.yml
jsotobroad: ah i see. i feel like before if a pipeline was done and `start` was called, it would return a fleshed out `AsyncPipelineRunResponse` but it looks like we no longer do that.  Hmm i wonder if we should make a ticket to have it either return the full response or return the `result` endpoint like it currently does when you call it successfullly.  Basically right now there is no way for this api to return a `200` so we could also just remove that.  All this for a different ticket, this is great as is!

mmorgantaylor: https://broadworkbench.atlassian.net/browse/TSPS-923

mmorgantaylor: for the record, calling start on a job that was running or done has always returned an error - this 200 response option probably should never have been here. so this isn't a breaking change.

MatthewBemis (approved):

PR #377, opened 2026-05-07
[TSPS-901] Update /status 503 response to reflect no object being returned

MatthewBemis:
### Description 

When /status returns a 503, it doesn't actually return an object like the Swagger was saying. This gets the swagger docs in line with reality.

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-901

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -733,12 +733,6 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/ErrorReport'
-    SystemStatusResponse:
-      description: A JSON description of the subsystems and their statuses.
-      content:
-        application/json:
-          schema:
-            $ref: '#/components/schemas/SystemStatus'
     JobResultResponse:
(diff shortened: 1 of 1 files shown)


common/openapi.yml
```diff
@@ -545,7 +545,7 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
         503:
-          $ref: '#/components/responses/SystemStatusResponse'
+          description: Service unavailable
```
salonishah11, line 548: So now the API will only return status code 503 and no response body?

MatthewBemis: The API already returned no body, this is just making the docs reflect that

jsotobroad (approved):

common/openapi.yml
salonishah11: thank you for the clarification

salonishah11 (approved):

PR #378, opened 2026-05-08
TSPS-851 Handle requester pays error messages

mmorgantaylor:
### Description 

When users submit data from an RP bucket (either the manifest/input file, or the files a manifeset points to), currently we return a generic "user doesn't have access to the bucket" message. Here we catch RP issues and return descriptive error messages to the user.

When file is in a RP bucket:
```
✗ terralab submit array_imputation --multiSampleVcf gs://sshah-testing/NA12878_10_samples_chr1_chr20.vcf.gz --outputBasename test --description "test RP bucket with file input"
Generated job_id ba59a69d-355b-4db4-bda6-69e398e57534

API call failed with status code 400: The bucket 'sshah-testing' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket.

For help troubleshooting, email scientific-services-support@broadinstitute.org
```

When manifest is in a RP bucket:
```
✗ terralab submit low_pass_imputation --cramManifest gs://sshah-testing/morgan-rp-testing/fakeCramManifestRPTargets.tsv --outputBasename test --description "test RP bucket"
Generated job_id d5b45bbd-3141-407d-b5e5-17622700dcd1

API call failed with status code 400: The bucket 'sshah-testing' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket.

For help troubleshooting, email scientific-services-support@broadinstitute.org
```

When manifest is ok but crams are in a manifest bucket:
```
✗ terralab submit low_pass_imputation --cramManifest gs://fc-secure-972057b6-fce1-4ec9-be0c-0e5c97fdc3e8/morgan-rp-testing/fakeCramManifestRPTargets.tsv --outputBasename test --description "test RP crams"
Generated job_id 0c924b9d-d0af-427a-91fc-a7b3943a5457

API call failed with status code 400: The bucket 'sshah-testing' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket.

For help troubleshooting, email scientific-services-support@broadinstitute.org
```

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
+++ b/service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
@@ -206,11 +206,21 @@ public class GcsService {
     return executionWithRetryTemplate(
         listenerResetRetryTemplate,
         () -> {
-          List<Boolean> accessResult =
-              gcsClient
-                  .getStorageService(accessToken)
-                  .testIamPermissions(bucketName, List.of("storage.objects.get"));
-          return accessResult.size() == 1 && accessResult.get(0);
+          try {
+            List<Boolean> accessResult =
+                gcsClient
+                    .getStorageService(accessToken)
+                    .testIamPermissions(bucketName, List.of("storage.objects.get"));
+            return accessResult.size() == 1 && accessResult.get(0);
+          } catch (StorageException e) {
+            if (e.getMessage().contains("Bucket is a requester pays bucket")) {
+              throw new RequesterPaysBucketException(
+                  "The bucket '%s' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket."
+                      .formatted(bucketName));
+            } else {
+              throw e;
+            }
+          }
         });
   }
 
--- a/service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
+++ b/service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
@@ -3,6 +3,7 @@ package bio.terra.pipelines.dependencies.gcs;
 import bio.terra.common.exception.InternalServerErrorException;
 import bio.terra.pipelines.app.configuration.external.GcsConfiguration;
 import bio.terra.pipelines.common.GcsFile;
+import bio.terra.pipelines.service.exception.RequesterPaysBucketException;
 import com.google.cloud.storage.*;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
@@ -84,6 +85,7 @@ public class GcsService {
    * @param gcsFile GcsFile object representing the GCS path to the file
    * @param accessToken for the calling user. Pass null to use application default credentials.
    * @return Blob object if the file exists and is accessible, null otherwise
+   * @throws RequesterPaysBucketException if the bucket is a requester pays bucket
    */
   public Blob getFileBlob(GcsFile gcsFile, String accessToken) {
     BlobId blobId = BlobId.fromGsUtilUri(gcsFile.getFullPath());
@@ -94,11 +96,17 @@ public class GcsService {
           .get(
               blobId, Storage.BlobGetOption.fields(Storage.BlobField.NAME, Storage.BlobField.SIZE));
     } catch (StorageException e) {
-      logger.error(
-          "An error occurred retrieving GCS file metadata for path `{}`. Error: {}",
-          gcsFile.getFullPath(),
-          e.getMessage());
-      return null;
+      if (e.getMessage().contains("Bucket is a requester pays bucket")) {
+        throw new RequesterPaysBucketException(
+            "The bucket for file '%s' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket."
+                .formatted(gcsFile.getFullPath()));
+      } else {
+        logger.error(
+            "An error occurred retrieving GCS file metadata for path `{}`. Error: {}",
+            gcsFile.getFullPath(),
+            e.getMessage());
+        return null;
+      }
     }
   }
 
@@ -137,6 +145,7 @@ public class GcsService {
    * @param gcsFilePath the full GCS path to the file (e.g. gs://my-bucket/path/to/file.txt)
    * @return the size of the file in bytes
    * @throws InternalServerErrorException if the file does not exist
+   * @throws RequesterPaysBucketException if the bucket is requester pays
    */
   public Long getFileSizeInBytes(String gcsFilePath) {
     GcsFile gcsFile = new GcsFile(gcsFilePath);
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
```diff
@@ -94,11 +96,17 @@ public Blob getFileBlob(GcsFile gcsFile, String accessToken) {
           .get(
               blobId, Storage.BlobGetOption.fields(Storage.BlobField.NAME, Storage.BlobField.SIZE));
     } catch (StorageException e) {
-      logger.error(
-          "An error occurred retrieving GCS file metadata for path `{}`. Error: {}",
-          gcsFile.getFullPath(),
-          e.getMessage());
-      return null;
+      if (e.getMessage().contains("Bucket is a requester pays bucket")) {
```
jsotobroad: hmm im guessing this is the best we can do with what information we have.  Like there isnt a field or anythign on `StorageException` that would let know its a requester pays thing so we just have to parse the error string (which they could change at any time) 🙄.

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
```diff
@@ -137,6 +145,7 @@ public BufferedReader getBufferedReaderForGcsTextFile(GcsFile gcsFile) {
    * @param gcsFilePath the full GCS path to the file (e.g. gs://my-bucket/path/to/file.txt)
    * @return the size of the file in bytes
    * @throws InternalServerErrorException if the file does not exist
+   * @throws RequesterPaysBucketException if the bucket is requester pays
```
jsotobroad: dont think you need to add this here as this function doesnt throw that exception explicitly

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
```diff
@@ -196,11 +206,21 @@ private boolean hasBucketReadAccess(String bucketName, String accessToken) {
...
+                    .getStorageService(accessToken)
+                    .testIamPermissions(bucketName, List.of("storage.objects.get"));
+            return accessResult.size() == 1 && accessResult.get(0);
+          } catch (StorageException e) {
+            if (e.getMessage().contains("Bucket is a requester pays bucket")) {
+              throw new RequesterPaysBucketException(
+                  "The bucket '%s' is a requester pays bucket, which is not currently supported. Please turn off requester pays or submit data from a non-requester pays bucket."
+                      .formatted(bucketName));
+            } else {
+              throw e;
```
jsotobroad: hmm why do we need to explicitly throw this whereas before we werent?  What was handling that exception before?

jsotobroad (commented):
can you add the example of a file input coming from a requester pays bucket to the description (just to cover all the cases)

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
mmorgantaylor: yeah unfortunately not that I could find, super annoying

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
mmorgantaylor: executionWithRetryTemplate catches StorageExceptions and converts them to GcsServiceExceptions. so we want this to be thrown if it's not a RP issue

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
mmorgantaylor: yeah I don't understand where these annotations are supposed to be used, but it seems helpful to point out that something inside here can throw this error so this method also can throw this error. if it's incorrect, i can take it out

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
```diff
@@ -158,18 +165,11 @@ public Long getFileSizeInBytes(String gcsFilePath) {
...
-      Blob blob = getFileBlob(gcsFile, accessToken);
-      if (blob != null && blob.exists()) {
-        return true;
-      }
-    } catch (StorageException e) {
-      logger.error("An error occurred checking blob access: {}", e.getMessage());
-      return false;
-    }
-    return false;
+    Blob blob = getFileBlob(gcsFile, accessToken);
```
mmorgantaylor, line 172: getFileBlob now handles catching exceptions and either throwing a RP error or returning null

service/src/main/java/bio/terra/pipelines/dependencies/gcs/GcsService.java
jsotobroad: yea handling "throws" is one of those things people write articles about.  LIke technically this method should have `throws InternalServerErrorException` in the signature since it throws that exception, and it should also have `throws RequesterPaysBucketException` because a method it calls should also have ` throws RequesterPaysBucketException` and then you just end up having that everywhere which can be good cuz then it makes sure you are handling all the exceptions you are throwing _somewhere_ . 

it looks like for spring boot a [standard (good?) practice](https://medium.com/@nayakasutosh176/exception-handling-in-spring-boot-a-complete-guide-for-production-applications-6bc5bf27ab64) is the approach we take with a global exception handler handlign all those exceptions so we dont need `throws` anywhere. yay us!

I do think its a little weird to have the `@throws` annotation without the `throws` signature field but i dont feel too strongly at all about this.

jsotobroad (approved):
i think adding an example for the requester pays file input would be great to add just to cover all the options. this is great

service/src/main/java/bio/terra/pipelines/service/exception/RequesterPaysBucketException.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.pipelines.service.exception;
+
+import bio.terra.common.exception.BadRequestException;
+
+@SuppressWarnings("java:S110") // Disable "Inheritance tree of classes should not be too deep"
```
MatthewBemis, line 5: lol i wonder if there's a way to do this at the project level since we have to do this for so many exception classes!

MatthewBemis (approved):

service/src/main/java/bio/terra/pipelines/service/exception/RequesterPaysBucketException.java
mmorgantaylor: ugh i know. it looks like we could specify certain issues repo-wide, though it seems like a good reminder to have to escape these deliberately, just to ask the question whether there's a way to do it differently. maybe something to bring up in our engineering discussion meeting?

PR #397, opened 2026-05-29
TSPS-912 pipeline definition refactor: MAJOR

mmorgantaylor:
### Description 

These changes are described at a high level in this Confluence page: https://broadworkbench.atlassian.net/wiki/spaces/TSPS/pages/4374953986/Pipeline+Config+YAML+Refactor+Summary+of+Changes+and+Architecture

To consolidate the source of truth for pipeline configuration, as well as simplify the input preparation logic in the service, we are refactoring the codebase to define a pipeline largely via YAML, keeping only the runtime-specific, admin-endpoint-controlled metadata in the codebase.

This work does not have any user-facing implications.

We did change the signature of the admin getPipeline response to add the `updated` field.

Testing:
- BEE e2e test: https://github.com/DataBiosphere/terra-scientific-pipelines-service/actions/runs/27017793201
- Local checks for the following:
  - as regular user:
    - list all pipelines
    - get pipeline details for all 3 pipelines (array imputation v1 and v2, low pass v2)
    - get all pipeline runs
    - get all pipeline runs with pipelineName filter
  - as admin user:
    - hide and unhide a pipeline via get/update pipeline admin endpoints 
    - list all pipelines - shows hidden pipeline
    - get pipeline - shows hidden pipeline
    - get & update user quota admin endpoints

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-912


service/src/main/java/bio/terra/pipelines/db/entities/PipelineRuntimeMetadata.java
```diff
@@ -0,0 +1,45 @@
...
+  @Column(name = "workspace_google_project")
+  private String workspaceGoogleProject;
+
+  @Column(name = "workspace_storage_container_name")
+  private String workspaceStorageContainerName;
+
+  @Column(name = "hidden", nullable = false)
+  private boolean hidden;
+
+  @Column(name = "created", insertable = false, updatable = false)
```
mmorgantaylor: todo: delete the `created` field (can keep `updated`)

rawls-client/swagger.gradle
```diff
@@ -10,15 +10,21 @@ dependencies {
 }
 
 def artifactGroup = "${group}.rawls"
+def rawlsCommit = "0ee920692b3c2173ed667e7219cecb0cc222a12a"
```
mmorgantaylor, line 13: drive-by update to avoid downloading this file every time you run a test

service/src/main/resources/pipelines-config.yml
```diff
@@ -28,23 +28,298 @@ pipelines:
...
+    pipelineQuotas:
+      array_imputation:
+        defaultQuota: 2500
+        minQuotaConsumed: 500
+        quotaUnits: SAMPLES
+      low_pass_imputation:
+        defaultQuota: 500
+        minQuotaConsumed: 500
+        quotaUnits: SAMPLES
     arrayImputation:
```
mmorgantaylor, line 40: in progress: changing this structure to be a bit more intutive (and avoid mixing cases of pipeline names)

rawls-client/swagger.gradle
```diff
@@ -10,15 +10,21 @@ dependencies {
...
+def cachedRawlsYaml =
+        file("${gradle.gradleUserHomeDir}/caches/teaspoons/rawls/rawls-api-docs-${rawlsCommit}.yaml")
```
mmorgantaylor: the robots suggested this location. happy to move it back to the repo so it gets cleaned up when we run `./gradlew clean` if people think that'd be better

service/src/main/java/bio/terra/pipelines/db/entities/PipelineRuntimeMetadata.java
```diff
@@ -0,0 +1,45 @@
...
+  private Instant created;
+
+  @Column(name = "updated", insertable = false, updatable = false)
+  private Instant updated;
```
mmorgantaylor, line 46: todo: admin get pipeline should return this field

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
```diff
@@ -0,0 +1,111 @@
...
+  // --- Display metadata (from PipelineDefinition) ---
+  private final String displayName;
+  private final String description;
+  private final String pipelineType;
+
+  // --- Definition structure (from PipelineDefinition) ---
+  private final String toolName;
+  private final List<PipelineInputDefinition> inputDefinitions;
+  private final List<PipelineOutputDefinition> outputDefinitions;
+  private final PipelineQuota quota;
```
mmorgantaylor: should this be removed from Pipeline?

service/src/main/resources/pipelines-config.yml
```diff
@@ -28,23 +28,299 @@ pipelines:
...
-        memoryRetryMultiplier: 2.0
-    lowPassImputation:
-      1:
-        inputKeysToPrependWithStorageWorkspaceContainerUrl: [ "refDict", "referencePanelPrefix", "fasta", "fastaIndex" ]
-        storageWorkspaceContainerUrl: ${env.pipelines.lowPassImputation.1.storageWorkspaceContainerUrl}
-        inputsWithCustomValues:
-          referencePanelPrefix: ${env.pipelines.lowPassImputation.1.referencePanelPrefixCustomValue}
-        memoryRetryMultiplier: 1.5
+    pipelineQuotas:
+      array_imputation:
```
mmorgantaylor: change these back to camelCase, add a `.getConfigKeyName` getter to PipelinesEnum to use for accessing these values

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -19,10 +28,239 @@
...
+
   private PipelinesCommonConfiguration common;
-  // this is a map of pipeline versions to their configurations
-  private Map<String, WdlBasedPipelineConfig> arrayImputation;
-  private Map<String, WdlBasedPipelineConfig> lowPassImputation;
+  private Map<String, PipelineQuotaConfig> pipelineQuotas;
+  private Map<String, Map<String, PipelineConfiguration>> pipelines;
+
+  @PostConstruct
+  public void validateConfiguration() {
```
mmorgantaylor: todo: move this validation to PipelineConfigurationsTest, pointing to service pipelines-config.yml

service/src/main/java/bio/terra/pipelines/service/pipeline/PipelineDefinitionProvider.java
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Get all pipeline definitions for a given pipeline name, sorted by version descending (highest
+   * first).
+   *
+   * @param name the pipeline name
+   * @return list of pipeline definitions for all versions, highest version first
+   */
+  public List<PipelineDefinition> getDefinitionsForPipeline(PipelinesEnum name) {
+    logger.debug("Getting all pipeline definitions for {}", name.getValue());
+    List<Integer> versions = getAvailableVersions(name);
```
mmorgantaylor: can just grab the version objects directly

rawls-client/swagger.gradle
jsotobroad: hmm would this file not actually get deleted if you run `clean` ?  I think i would definitely want it to be deleted with a clean.  If that's not possible I would rather download the file each time than have to remember to manually delete the file when trying to pull in a new commit of the yaml

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -19,10 +28,239 @@
 @ConfigurationProperties(prefix = "pipelines.configurations")
 public class PipelineConfigurations {
 
+  private static final Pattern PIPELINE_KEY_PATTERN = Pattern.compile("^[a-z0-9_]+_v\\d+$");
+
   private PipelinesCommonConfiguration common;
-  // this is a map of pipeline versions to their configurations
-  private Map<String, WdlBasedPipelineConfig> arrayImputation;
```
jsotobroad, line 24: is there a reason we got rid of the `WdlBasedPipelineConfig` name for this refactor?

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -19,10 +28,239 @@
...
+      if (input == null) {
+        throw new IllegalArgumentException(
+            "Input definition cannot be null for pipeline '%s'".formatted(pipelineKey));
+      }
+
+      String inputName = requireText(input.getName(), "inputs.name", pipelineKey);
+      requireText(input.getWdlVariableName(), "inputs.wdlVariableName", pipelineKey);
+      requireNonNull(input.getType(), "inputs.type", pipelineKey);
+      requireNonNull(input.getIsRequired(), "inputs.isRequired", pipelineKey);
+      requireNonNull(input.getUserProvided(), "inputs.userProvided", pipelineKey);
```
jsotobroad: same question about about `requireNonNull` vs `requireText` for these fields

I might look through all of these and double check

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -19,10 +28,239 @@
...
+    }
+    PipelineQuotaConfig quota = pipelineQuotas.get(pipelineName.getValue());
+    if (quota == null) {
+      throw new IllegalArgumentException(
+          "No quota configuration found for pipeline '%s'".formatted(pipelineName.getValue()));
+    }
+    return quota;
+  }
+
+  private String requireText(String value, String fieldName, String pipelineKey) {
```
jsotobroad: can we name this to `requireNonEmpty`?  at least to me that ends up making more sense

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -19,10 +28,239 @@
...
+    if (value == null || value.isBlank()) {
+      throw new IllegalArgumentException(
+          "Missing required field '%s' for pipeline '%s'".formatted(fieldName, pipelineKey));
+    }
+    return value;
+  }
+
+  private <T> T requireNonNull(T value, String fieldName, String pipelineKey) {
```
jsotobroad: oh i see we seem to use this when the type of the field being checked is not a string.  well sort of this is also used for some string type fields like `inputs.type` or `output.type`

I think if we want it to be equivalent to what was happneing in the database.  we woudl want a validation check for each type we expect like `requireBoolean`, `requireInt`, `requireFloat`, etc, etc.

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -98,9 +98,10 @@ static ApiGetPipelinesResult pipelinesToApi(List<Pipeline> pipelineList) {
     return apiResult;
   }
 
-  static ApiPipelineQuota pipelineQuotaToApi(PipelineQuota pipelineQuota) {
+  static ApiPipelineQuota pipelineQuotaToApi(
+      PipelinesEnum pipelineName, PipelineQuota pipelineQuota) {
```
jsotobroad: is it possible for the PipelineQuota object having the pipeline name as one of its fields?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineRunFilterSpecification.java
```diff
@@ -116,14 +110,8 @@ private static Predicate validateAndBuildPipelineNamePredicate(
...
-    // Use a subquery to find the Pipeline id(s) matching the given name,
-    // since PipelineRun only stores pipelineId (no JPA relationship to Pipeline).
-    Subquery<Long> subquery = query.subquery(Long.class);
-    Root<Pipeline> pipelineRoot = subquery.from(Pipeline.class);
-    subquery
-        .select(pipelineRoot.get("id"))
-        .where(criteriaBuilder.equal(pipelineRoot.get("name"), pipelineName));
-    return root.get("pipelineId").in(subquery);
+    // Match canonical pipeline keys (e.g. array_imputation_v2) by pipeline name prefix.
+    return criteriaBuilder.like(root.get("pipelineKey"), pipelineName.getValue() + "_v%");
```
jsotobroad: ooooo this looks much simpler!  what does the `%` in `"_v%"` do?  Im not a huge fan of `like`s in queries unless its 100% necessary .  I dont think I know enough of what is going on to know one way or the other.  Happy to chat!

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -87,7 +87,7 @@ public ResponseEntity<ApiAdminQuota> getQuotaForPipelineAndUser(
       throw new BadRequestException(
           String.format(
               "User quota not found for user %s and pipeline %s",
-              userId, validatedPipelineName.getValue()));
+              userId, validatedPipelineName.getLowerCaseValue()));
```
jsotobroad, line 90: is this a drive by fix to make look less bad?

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -159,7 +159,7 @@ public ResponseEntity<ApiAdminQuota> updateQuotaLimitForPipelineAndUser(
 
   public ApiAdminPipeline pipelineToApiAdminPipeline(Pipeline pipeline) {
     return new ApiAdminPipeline()
-        .pipelineName(pipeline.getName().getValue())
+        .pipelineName(pipeline.getName().getLowerCaseValue())
```
jsotobroad, line 162: hmm why does this need to be lower cased?  This seems to be needed in a lot of new places now

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -500,7 +495,10 @@ private ApiAsyncPipelineRunResponseV2 pipelineRunToApiV2(
 
     Integer inputSize = pipelineRun.getRawQuotaConsumed();
     if (inputSize != null) {
-      String inputSizeUnits = quotasService.getQuotaUnitsForPipeline(pipeline.getName()).getValue();
+      String inputSizeUnits =
+          quotasService
+              .getQuotaUnitsForPipeline(PipelinesEnum.enumFromPipelineKey(pipelineKey))
```
jsotobroad: should this function be named to `getQuotaUnitsForPipelineKey()` ?

service/src/main/java/bio/terra/pipelines/common/utils/PipelinesEnum.java
```diff
@@ -1,16 +1,99 @@
...
+  private static final String PIPELINE_NOT_FOUND_MESSAGE = "Pipeline not found for pipelineKey %s";
+
+  /**
+   * Build the canonical pipeline key from a name enum and integer version.
+   *
+   * @param name the pipeline name enum
+   * @param version the pipeline version
+   * @return key in the form {@code array_imputation_v2}
+   */
+  public static String buildPipelineKey(PipelinesEnum name, int version) {
```
jsotobroad: Im not sure these pipeline key related functions/fields make sense to have in this enum class other than the ones that explictly convert from a value to an enum.  Maybe a `PipelineConfigHelper` or something

service/src/main/java/bio/terra/pipelines/db/repositories/PipelineRuntimeMetadataRepository.java
```diff
@@ -0,0 +1,22 @@
...
+  @Override
+  List<PipelineRuntimeMetadata> findAll();
+
+  /**
+   * Find all runtime metadata entries for a given pipeline name.
+   *
+   * @param pipelineKeyPrefix the pipeline name prefix (e.g., "array_imputation_")
+   * @return list of metadata entries matching the prefix
+   */
+  List<PipelineRuntimeMetadata> findAllByPipelineKeyStartingWith(String pipelineKeyPrefix);
```
jsotobroad: hmm is there any way we can just put pipeline name and version in this table too?  I just feels a little odd to have these types of functions be our main way of finding rows.

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
```diff
@@ -0,0 +1,112 @@
...
+            .displayName(definition.getDisplayName())
+            .description(definition.getDescription())
+            .pipelineType(definition.getPipelineType())
+            .toolName(definition.getToolName())
+            .inputDefinitions(definition.getInputDefinitions())
+            .outputDefinitions(definition.getOutputDefinitions())
+            .memoryRetryMultiplier(definition.getMemoryRetryMultiplier())
+            .hidden(true);
+
+    if (runtimeMetadata != null) {
```
jsotobroad: im still not sure i even want to be able to get a pipeline that doesnt have associated runtimeMetadata row 🤔

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
```diff
@@ -28,6 +28,6 @@ public class JobMapKeys {
   }
 
   public static List<String> getRequiredKeys() {
-    return Arrays.asList(USER_ID, PIPELINE_NAME, PIPELINE_ID);
+    return Arrays.asList(USER_ID, PIPELINE_NAME, PIPELINE_KEY);
```
jsotobroad: is `PIPELINE_NAME` still necessary if we have `PIPELINE_KEY` now?

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
```diff
@@ -0,0 +1,23 @@
...
+@ToString
+@EqualsAndHashCode
+@NoArgsConstructor(force = true)
+@SuperBuilder(toBuilder = true)
+public abstract class BasePipelineVariableDefinition {
+  private final String name;
+  private final String wdlVariableName;
+  private final String displayName;
+  private final String description;
+  private final PipelineVariableTypesEnum type;
```
jsotobroad: I know when this was in teh database it required a convert  to go from string to enum value.  Does this now happen automagically as part of the config parsing process?

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Immutable domain model representing a pipeline as seen by the service and controller layers. This
+ * combines pipeline definition content (from YAML via PipelineDefinition) with runtime metadata
+ * (from the pipeline_runtime_metadata table).
+ */
+@Getter
+@ToString
+@EqualsAndHashCode
+@Builder(toBuilder = true)
+public class Pipeline {
```
jsotobroad, line 21: is it actually necessary to have this representation of the yaml stuff + runtime table?  Like do we always need both at all times?  The answer could be yes I'm just not sure.

jsotobroad: also if we do keep this class can it extend `PipelineDefinition`?

service/src/main/java/bio/terra/pipelines/service/pipeline/PipelineDefinitionProvider.java
```diff
@@ -0,0 +1,190 @@
...
+        .build();
+  }
+
+  /**
+   * Transform input definition configs into domain models.
+   *
+   * @param inputConfigs the configuration objects
+   * @return list of domain model input definitions
+   */
+  private List<PipelineInputDefinition> transformInputs(
```
jsotobroad: `transformInputDefinitions`

service/src/main/java/bio/terra/pipelines/service/pipeline/PipelineDefinitionProvider.java
```diff
@@ -0,0 +1,190 @@
...
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Transform output definition configs into domain models.
+   *
+   * @param outputConfigs the configuration objects
+   * @return list of domain model output definitions
+   */
+  private List<PipelineOutputDefinition> transformOutputs(
```
jsotobroad: `transformOutputDefinitions`

service/src/main/java/bio/terra/pipelines/service/pipeline/PipelineDefinitionProvider.java
```diff
@@ -0,0 +1,190 @@
...
+   * @param version the pipeline version
+   * @return the pipeline definition
+   * @throws NotFoundException if the pipeline is not found in the YAML configuration
+   */
+  public PipelineDefinition getPipelineDefinition(PipelinesEnum name, Integer version) {
+    logger.debug("Getting pipeline definition for {} v{}", name.getLowerCaseValue(), version);
+    String pipelineKey = PipelinesEnum.buildPipelineKey(name, version);
+    PipelineConfigurations.PipelineConfiguration wdlConfig =
+        pipelineConfigurations.getPipelineConfiguration(pipelineKey);
+    return transformToDomainModel(wdlConfig, name, version, pipelineKey);
```
jsotobroad: does this mean we do a transformation from the yaml each time we ask for a particular pipeline?  Would it be possible to only have to do this once per instantiating of the service per pipeline?

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
```diff
@@ -809,38 +810,27 @@ Map<String, Object> formatPipelineInputs(
...
 
-      if (keysToPrependWithStorageWorkspaceContainerUrl.contains(keyName)) {
-        // the rawOrCustomValue for this field should start with a / so we don't need to add one
-        // here
-        processedValue = constructFilePath(storageWorkspaceContainerUrl, rawOrCustomValue);
-      } else if (inputDefinition.isUserProvided()
-          && inputDefinition.getType().equals(PipelineVariableTypesEnum.FILE)
-          && getFileLocationType(rawOrCustomValue) == FileLocationTypeEnum.LOCAL) {
+      if (inputDefinition.isUserProvided()
+          && inputDefinition.getType().isFileLike()
```
jsotobroad, line 823: is this a drive by fix?

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
```diff
@@ -1074,14 +1040,17 @@ public Map<String, Object> stringToMap(String outputsString) {
...
    *
-   * @param pipeline the pipeline to get the file outputs for
+   * @param pipelineKey the key for the pipeline to get the file outputs for
    * @param outputsMap a map of the pipeline outputs
    * @return a map with output name and their corresponding file sizes in bytes
    */
-  public Map<String, Long> getPipelineOutputsFileSize(
-      Pipeline pipeline, Map<String, String> outputsMap) {
+  public Map<String, Long> getPipelineOutputsFileSizeByPipelineKey(
+      String pipelineKey, Map<String, String> outputsMap) {
```
jsotobroad, line 1046: is there a reason we cant pass around the new Pipeline model class like we did the Pipeline entity class?  Is there a clear distinction for where in the code do we expect to call the pipeline configuration directly vs passing around a pipeline object?

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -174,6 +171,7 @@ public PipelineRun startPipelineRun(Pipeline pipeline, UUID jobId, SamUser authe
     String userId = authedUser.getSubjectId();
 
     validatePipelineWorkspaceSetup(pipeline);
+    validatePipelineKey(pipeline.getPipelineKey(), pipeline.getName().toString());
```
jsotobroad: is this necessary if the pipeline config has this column as non nullable?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -33,32 +38,56 @@ public class PipelinesService {
...
    *     for admin users)
    * @return List of pipelines ordered by name (alphabetically) and version (descending)
    */
   public List<Pipeline> getPipelines(boolean showHidden) {
     logger.info("Get all Pipelines");
-    if (showHidden) {
-      return pipelinesRepository.findAllByOrderByNameAscVersionDesc();
+    List<Pipeline> pipelines = new ArrayList<>();
+
+    for (PipelinesEnum pipelineEnum : PipelinesEnum.values()) {
```
jsotobroad: would it be possible to go through the pipeline metadata first, and then for the ones that pass go grab their pipelinedefinition ?  This is jsut because the metadata table will at most have as many passing entities as the yaml.

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -33,32 +38,56 @@ public class PipelinesService {
...
+      List<PipelineDefinition> definitions =
+          pipelineDefinitionProvider.getDefinitionsForPipeline(pipelineEnum);
+      for (PipelineDefinition definition : definitions) {
+        PipelineRuntimeMetadata runtimeMetadata =
+            pipelineRuntimeMetadataRepository.findById(definition.getPipelineKey()).orElse(null);
+
+        if (runtimeMetadata == null || (runtimeMetadata.isHidden() && !showHidden)) {
```
jsotobroad: can this be written the same way as the same cheeck is done in the `getPipeline` function

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -72,57 +101,93 @@ public Pipeline getPipeline(
...
 
-  public Pipeline getLatestPipeline(PipelinesEnum pipelineName) {
+  /**
+   * Get the latest version of a pipeline.
+   *
+   * @param pipelineName name of the pipeline
+   * @param showHidden whether to include hidden pipelines
+   * @return the latest pipeline
+   */
+  private Pipeline getLatestPipeline(PipelinesEnum pipelineName, boolean showHidden) {
```
jsotobroad: whats the reasoning for this change in the function signature? is it something we're making use of now?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -72,57 +101,93 @@ public Pipeline getPipeline(
...
-    Pipeline dbResult =
-        pipelinesRepository.findFirstByNameAndHiddenIsFalseOrderByVersionDesc(pipelineName);
-    if (dbResult == null) {
+
+    // Query runtime metadata for all versions of this pipeline
+    String pipelineKeyPrefix = pipelineName.getLowerCaseValue() + "_v";
+    List<PipelineRuntimeMetadata> metadataList =
+        pipelineRuntimeMetadataRepository.findAllByPipelineKeyStartingWith(pipelineKeyPrefix);
+
+    if (metadataList.isEmpty()) {
```
jsotobroad: i assume you tested this but would this be an empty list or null?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -72,57 +101,93 @@ public Pipeline getPipeline(
...
+  private Pipeline getLatestPipeline(PipelinesEnum pipelineName, boolean showHidden) {
     logger.info("Get the latest pipeline for pipelineName {}", pipelineName);
-    Pipeline dbResult =
-        pipelinesRepository.findFirstByNameAndHiddenIsFalseOrderByVersionDesc(pipelineName);
-    if (dbResult == null) {
+
+    // Query runtime metadata for all versions of this pipeline
+    String pipelineKeyPrefix = pipelineName.getLowerCaseValue() + "_v";
+    List<PipelineRuntimeMetadata> metadataList =
+        pipelineRuntimeMetadataRepository.findAllByPipelineKeyStartingWith(pipelineKeyPrefix);
```
jsotobroad: why cant you use a similar `findFirstByNameAndHiddenIsFalseOrderByVersionDesc` type of call for this?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -72,57 +101,93 @@ public Pipeline getPipeline(
...
-   * as control workspace information and wdl method version.
+   * Resolve a pipeline by canonical pipelineKey ({pipeline_name}_v{version}).
    *
-   * <p>Calls Rawls to fetch control workspace metadata based on the workspaceBillingProject and
-   * workspaceName.
+   * @param pipelineKey canonical pipeline key
+   * @param showHidden whether to include hidden pipelines
+   * @return merged pipeline definition/runtime metadata
+   */
+  public Pipeline getPipelineByKey(String pipelineKey, boolean showHidden) {
```
jsotobroad: `getPipelineByPipelineKey()`

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+/** Shared immutable fields for pipeline input/output definitions in the domain model. */
+@Getter
+@ToString
+@EqualsAndHashCode
+@NoArgsConstructor(force = true)
+@SuperBuilder(toBuilder = true)
```
jsotobroad: what do these `@Builder` and `@SuperBuilder` decorators do?

service/src/main/java/bio/terra/pipelines/service/ToolConfigService.java
```diff
@@ -26,34 +25,24 @@ public ToolConfigService(PipelineConfigurations pipelineConfigurations) {
...
         getDataTableEntityNameForToolConfig(pipeline),
-        pipeline.getPipelineInputDefinitions(),
-        pipeline.getPipelineOutputDefinitions(),
+        pipeline.getInputDefinitions(),
+        pipeline.getOutputDefinitions(),
         commonConfiguration.isMainToolUseCallCaching(),
         commonConfiguration.getMonitoringScriptPath(),
         commonConfiguration.isMainToolDeleteIntermediateFiles(),
-        wdlBasedPipelineConfig.getMemoryRetryMultiplier(),
+        metadata == null ? null : metadata.getMemoryRetryMultiplier(),
```
jsotobroad: is metadata existing not a requirement for our pipelines?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -155,9 +212,29 @@ public Pipeline adminUpdatePipelineWorkspace(
               "toolVersion %s does not follow semantic versioning regex %s",
               toolVersion, SEM_VER_REGEX_STRING));
     }
-    pipeline.setToolVersion(toolVersion);
 
-    pipelinesRepository.save(pipeline);
-    return pipeline;
+    // Build an updated model with the new runtime values
+    Pipeline updatedPipeline =
```
jsotobroad: is this done just to return an up to date representation of the pipeline or is this actually being "stored" somewhere?

service/src/main/java/bio/terra/pipelines/stairway/flights/wdlbasedpipelinerun/v20260428/RunWdlBasedPipelineJobFlight.java
```diff
@@ -1,7 +1,6 @@
 package bio.terra.pipelines.stairway.flights.wdlbasedpipelinerun.v20260428;
```
jsotobroad: we should update the date here just so its obvious its meant to be a differnet version of the flight

service/src/main/java/bio/terra/pipelines/stairway/steps/common/PopulateFileOutputSizeStep.java
```diff
@@ -43,8 +37,7 @@ public StepResult doStep(FlightContext flightContext) {
     try {
       // validate and extract parameters from input map
       var inputParameters = flightContext.getInputParameters();
-      FlightUtils.validateRequiredEntries(inputParameters, JobMapKeys.PIPELINE_ID);
-      Long pipelineId = inputParameters.get(JobMapKeys.PIPELINE_ID, Long.class);
+      FlightUtils.validateRequiredEntries(inputParameters, JobMapKeys.PIPELINE_KEY);
```
jsotobroad, line 40: update step java docs please (same for every step where this was changed)

service/src/main/java/bio/terra/pipelines/stairway/steps/common/PrepareInputsStep.java
```diff
@@ -53,7 +48,8 @@ public StepResult doStep(FlightContext flightContext) {
         WdlBasedPipelineJobMapKeys.CONTROL_WORKSPACE_STORAGE_CONTAINER_NAME);
 
     PipelinesEnum pipelineEnum =
-        PipelinesEnum.valueOf(inputParameters.get(JobMapKeys.PIPELINE_NAME, String.class));
+        PipelinesEnum.enumFromLowerCaseValue(
```
jsotobroad: it seems like we're alternating between upper case and lowercase representations of the pipeline name.  is there a way to make this more consistent?

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
jsotobroad, line 807: it'll be nice to clean up the java doc for this function :D

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
```diff
@@ -0,0 +1,142 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-pipeline-key-columns
+      author: copilot
```
jsotobroad: not sure i love copilot as the author hahaha

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
```diff
@@ -0,0 +1,142 @@
...
+              FROM pipelines p
+              WHERE pr.pipeline_id = p.id
+                AND pr.pipeline_key IS NULL;
+
+  - changeSet:
+      id: add-pipeline-key-constraints-and-indexes
+      author: copilot
+      changes:
+        - addNotNullConstraint:
+            tableName: pipelines
```
jsotobroad: is ther eany need to update the pipelines table in this migration?

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+  - changeSet:
+      id: backfill-pipeline-key-and-create-runtime-metadata
+      author: copilot
+      changes:
+        - sql:
+            dbms: postgresql
+            sql: >
+              UPDATE pipelines
+              SET pipeline_key = lower(name::text) || '_v' || version::text
```
jsotobroad: oh is this where all the lower case transformations are accommodating?  maybe we can change the enum value to be lower case instead of upper case

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
```diff
@@ -0,0 +1,142 @@
...
+        - addUniqueConstraint:
+            tableName: pipelines
+            columnNames: pipeline_key
+            constraintName: pipelines_pipeline_key_uk
+        - addNotNullConstraint:
+            tableName: pipeline_runs
+            columnName: pipeline_key
+            columnDataType: text
+        - addForeignKeyConstraint:
+            baseTableName: pipeline_runs
```
jsotobroad, line 125: im guessing we'll need a ticket to drop the pipeline_id column and the pipeline table in the future?

service/src/main/resources/pipelines-config.yml
```diff
@@ -28,23 +28,299 @@ pipelines:
...
+    pipelineQuotas:
+      arrayImputation:
+        defaultQuota: 2500
+        minQuotaConsumed: 500
+        quotaUnits: SAMPLES
+      lowPassImputation:
+        defaultQuota: 500
+        minQuotaConsumed: 500
+        quotaUnits: SAMPLES
+    pipelines:
```
jsotobroad: it is a little strange to have this yaml path be `pipelines.configurations.pipelines`.

jsotobroad (commented):
happy to chat about these comments :D

rawls-client/swagger.gradle
mmorgantaylor: i could not actually find this file anywhere. so i'm putting it back where it was (rawls-client/build/). i confirmed it gets cleaned with clean. leaving the hash in the downloaded file name.

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
mmorgantaylor: it's to make it more clear what "value" is - now that we have both lowerCaseValue and configKeyValue

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
mmorgantaylor: shouldn't be any new places. it's just renamed.

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
mmorgantaylor: no, `PipelinesEnum.enumFromPipelineKey(pipelineKey)` resolves to a PipelinesEnum

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
mmorgantaylor: ah yes it does, good call. what a dumb change, thanks for catching it

service/src/main/java/bio/terra/pipelines/common/utils/PipelinesEnum.java
mmorgantaylor: yeahhhh I agree. the robots put them here and it seemed fine but not great. a helper sounds good

service/src/main/java/bio/terra/pipelines/dependencies/stairway/JobMapKeys.java
mmorgantaylor: it's not - I was tempted to revisit what we pass to flights - we can likely just keep PIPELINE_KEY - but didn't want to expand scope. a number of steps do use PIPELINE_NAME so removing that here will force the larger refactor. let's chat - I'd love to do it if folks are ok with it.

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
mmorgantaylor: they enable using `.builder()` to build instances of the class rather than a constructor.

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
mmorgantaylor: yeah! in PipelineConfigurations classes for input and output definitions, that field is specified from the start as the enum:
`private PipelineVariableTypesEnum type;`

if we provide an invalid type in the yaml, everything breaks - here's what the error says when running tests with `MADEUPTYPE` as one of the input definition types: 
```
Caused by: org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [bio.terra.pipelines.common.utils.PipelineVariableTypesEnum] for value [MADEUPTYPE]
```

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
mmorgantaylor: we often don't need both, and I think we could break them up if we want - the idea here was to keep the "Pipeline" interface that services are used to consistent. I'll see what it looks like to separate them. I agree it doesn't feel great to do it this way.

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
mmorgantaylor: not allowing this would be a change from current behavior - currently if the runtime metadata fields in `pipelines` aren't specified (which they aren't at startup for a new pipeline or a fresh environment), you can still see the pipeline, just those values aren't populated, and it only fails when trying to prepare a job. 

I do prefer the alternative behavior implied by your suggestion - that even getting a pipeline in this case would return an error. just calling out that it's a change of user-facing behavior. happy to chat and decide if we want to do this.

service/src/main/java/bio/terra/pipelines/service/pipeline/PipelineDefinitionProvider.java
mmorgantaylor: ah ok this is what the cache was doing. putting that back, now I understand it.

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
mmorgantaylor: yes, though i'm confused how this was working with local manifest uploads... maybe it wasn't? i don't think we have an e2e test for that yet

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
mmorgantaylor: there's definitely not a clear distinction in the code for when to pass around the object vs the key. we should chat about this.

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
mmorgantaylor: good catch, not needed!

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: yeah, that makes sense

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: actually with querying the metadata table first this will never return pipelines without a row in the metadata table. i think this is ok, it's a change in behavior, we can chat about it

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: hmm good question. i think i was worried about admins being able to get hte latest pipeline, but we require admins to specify the version, so this shouldn't be called from admin getPipeline ever. the other place it's used is from quota stuff, just to pull the latest display name. so you're right, this never has to be true. I'll change it back.

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: we could if we had version in the table.

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
mmorgantaylor: we can chat about this

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: it's just to return the representation. so the API can use it. this will probably get refactored

service/src/main/java/bio/terra/pipelines/service/ToolConfigService.java
mmorgantaylor: yeah this was a robot getting confused, thanks for catching

service/src/main/java/bio/terra/pipelines/stairway/flights/wdlbasedpipelinerun/v20260428/RunWdlBasedPipelineJobFlight.java
mmorgantaylor: yes, I plan to - I just want to wait till the last minute to actually do this so that the changes are easier to review. thanks for calling out, I'll add a TODO so I don't forget

mmorgantaylor: oh actually renaming it entirely (instead of making a copy) will be fine for reviewing. good call!

service/src/main/java/bio/terra/pipelines/stairway/steps/common/PrepareInputsStep.java
mmorgantaylor: we're not really - this is the same behavior as before. I think this will get cleaner if we decide to just use PIPELINE_KEY in the map

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
mmorgantaylor: lol fair i will sign my name

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
mmorgantaylor: yep, it exists: https://broadworkbench.atlassian.net/browse/TSPS-963

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
mmorgantaylor: nope

service/src/main/resources/pipelines-config.yml
mmorgantaylor: i agree. please suggest how to improve it. i spent too much time trying to make this yaml intutive

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
mmorgantaylor: we can make enum values lower case and this would simplify a lot of this stuff, yes - it would go against Java naming conventions (source: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) but i'm ok with it if the team is

service/src/main/resources/pipelines-config.yml
mmorgantaylor: i'm trying out removing the top-level pipelines entirely, so just starting with configurations

mmorgantaylor: seems to work

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
mmorgantaylor: breaking this up will be a pretty big refactor the way we use the combination of these two sources. we can consider doing it but we should chat.

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
mmorgantaylor: reverted the rename

docs/pipeline-config-yaml-refactor-plan.MD
```diff
@@ -0,0 +1,183 @@
+# Pipeline Config YAML Refactor Plan
```
MatthewBemis: thanks for including this! is it something we're going to keep around or just for the duration of review? some of it is a bit out of date with the recent updates

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -16,17 +28,18 @@
...
+  private static final Pattern PIPELINE_KEY_PATTERN = Pattern.compile("^[a-z0-9_]+_v\\d+$");
+
+  private CommonConfiguration common;
+  private Map<String, PipelineQuotaConfiguration> pipelineQuotas;
+  private Map<String, Map<String, WdlBasedPipelineConfiguration>> pipelines;
 
   @Getter
   @Setter
-  public static class PipelinesCommonConfiguration {
+  public static class CommonConfiguration {
```
MatthewBemis: sorry if this was discussed and i'm forgetting- but what's the justification for this rename? `CommonConfiguration` feels possibly a bit too generic/vague

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
```diff
@@ -68,26 +69,26 @@ public class PipelineInputsOutputsService {
...
+    this.pipelineConfigurations = pipelineConfigurations;
   }
 
   /**
    * Helper function to get FILE type (including subset types, i.e. MANIFEST) input keys that are
    * user-provided for a pipeline
    */
   private List<String> getUserProvidedFileInputKeys(Pipeline pipeline) {
-    return pipeline.getPipelineInputDefinitions().stream()
+    return pipeline.getInputDefinitions().stream()
```
MatthewBemis, line 91: i like this shorter rename

service/src/test/java/bio/terra/pipelines/dependencies/rawls/RawlsServiceTest.java
```diff
@@ -461,12 +461,11 @@ void checkHealthWithException() throws ApiException {
 
   private PipelineInputDefinition generatePipelineInputDefinitionWithWdlVariableName(
       String wdlVariableName) {
-    return new PipelineInputDefinition(
-        null, null, wdlVariableName, null, null, null, null, true, true, false, null, null, null);
+    return PipelineInputDefinition.builder().wdlVariableName(wdlVariableName).build();
```
MatthewBemis, line 464: nice cleanup

README.md
```diff
@@ -78,6 +78,48 @@ To run locally:
       4. workspaceName is also listed in the Teaspoons Resources document, and can be found through the Terra UI workspace dashboard
       5. wdlMethodVersion is found for the specific workflow as listed in the Terra UI page for workflows.
 
+#### Back up local Postgres databases before testing/refactors
```
MatthewBemis, line 81: nice!

service/src/main/java/bio/terra/pipelines/db/repositories/PipelineRuntimeMetadataRepository.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  List<PipelineRuntimeMetadata> findAll();
+
+  /**
+   * Find the latest non-hidden runtime metadata entry for a given pipeline name
+   *
+   * @param pipelineKeyPrefix the pipeline name prefix (e.g., "array_imputation")
+   * @return list of metadata entries matching the prefix
+   */
+  PipelineRuntimeMetadata findFirstByPipelineKeyStartingWithAndHiddenIsFalseOrderByPipelineKeyDesc(
```
MatthewBemis: still in awe every time i see one of these crazy names that _just works_

docs/pipeline-config-yaml-refactor-plan.MD
mmorgantaylor: yeah I've revised it a bit and plan to turn it into more of a summary. we can keep it if we think it's useful, or get rid of it.

service/src/main/java/bio/terra/pipelines/db/repositories/PipelineRuntimeMetadataRepository.java
mmorgantaylor: lol me too

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
mmorgantaylor: nah, that was a spur-of-the-moment choice by me but happy to revert it. you're right that CommonConfiguration is too generic. thanks!

service/src/main/java/bio/terra/pipelines/app/configuration/internal/PipelineConfigurations.java
```diff
@@ -40,31 +47,90 @@ public static class PipelinesCommonConfiguration {
...
+  public static class PipelineQuotaConfiguration {
+    private Integer defaultQuota;
+    private Integer minQuotaConsumed;
+    private QuotaUnitsEnum quotaUnits;
+  }
+
+  public PipelineQuotaConfiguration getQuotaForPipeline(PipelinesEnum pipelineName) {
+    if (pipelineQuotas == null) {
+      throw new IllegalArgumentException(
+          "pipelines.configurations.pipelineQuotas is not configured");
```
jsotobroad: update path (removed `pipelines.`)

common/openapi.yml
```diff
@@ -821,6 +821,8 @@ components:
           $ref: "#/components/schemas/PipelineWorkspaceStorageContainerName"
         workspaceGoogleProject:
           $ref: "#/components/schemas/PipelineWorkspaceGoogleProject"
+        updated:
```
jsotobroad, line 824: nice thank you for this update!

service/src/main/java/bio/terra/pipelines/app/controller/PipelineApiUtils.java
```diff
@@ -22,7 +22,7 @@ private PipelineApiUtils() {}
    */
   public static PipelinesEnum validatePipelineName(String pipelineName, Logger logger) {
     try {
-      return PipelinesEnum.valueOf(pipelineName.toUpperCase());
+      return PipelinesEnum.enumFromLowerCaseValue(pipelineName);
```
jsotobroad: what is driving the change from uppercase things to lower case things?  Its ok im just curious what the reason is

Also i think we woudl want to lower case `pipelineName` in this function if we're going to be grabbing the enum this way.  I dont think its guaranteed `pipelineName` is lowercase since I believe user input can be passed to this function

service/src/main/java/bio/terra/pipelines/common/utils/PipelineRunFilterSpecification.java
```diff
@@ -116,14 +110,8 @@ private static Predicate validateAndBuildPipelineNamePredicate(
...
-    // Use a subquery to find the Pipeline id(s) matching the given name,
-    // since PipelineRun only stores pipelineId (no JPA relationship to Pipeline).
-    Subquery<Long> subquery = query.subquery(Long.class);
-    Root<Pipeline> pipelineRoot = subquery.from(Pipeline.class);
-    subquery
-        .select(pipelineRoot.get("id"))
-        .where(criteriaBuilder.equal(pipelineRoot.get("name"), pipelineName));
-    return root.get("pipelineId").in(subquery);
+    // Match canonical pipeline keys (e.g. array_imputation_v2) by pipeline name prefix.
+    return criteriaBuilder.like(root.get("pipelineKey"), pipelineName.getLowerCaseValue() + "_v%");
```
jsotobroad: this look good but i dont think we have any live tests for this.  Can we just make sure this still works (we could also make a ticket to add a test of this to the live e2e test maybe?

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
jsotobroad: nice

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
jsotobroad: 👍 as long as its impossible of having "stale" runtime metadata this is good

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
```diff
@@ -0,0 +1,68 @@
...
+
+  // --- Identity (config) ---
+  private final PipelinesEnum name;
```
jsotobroad, line 24: are we still planning on replacing the "config" fields in this class with a PipelineDefinitions field?

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -33,32 +49,110 @@ public class PipelinesService {
...
+      PipelineRuntimeMetadataRepository pipelineRuntimeMetadataRepository,
+      RawlsService rawlsService,
+      SamService samService) {
+    this.pipelineConfigurations = pipelineConfigurations;
+    this.pipelineRuntimeMetadataRepository = pipelineRuntimeMetadataRepository;
     this.rawlsService = rawlsService;
     this.samService = samService;
   }
 
+  private List<PipelineInputDefinition> inputDefinitionsFromConfig(
```
jsotobroad: hmm what is the difference between a `PipelineInputDefinitionConfiguration` and a `PipelineInputDefinition`? Same question for the output equivalent below

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
```diff
@@ -155,9 +276,50 @@ public Pipeline adminUpdatePipelineWorkspace(
...
+            .orElse(new PipelineRuntimeMetadata());
+
+    // if it's a new object, populate the key.
+    if (runtimeMetadata.getPipelineKey() == null) {
+      runtimeMetadata.setPipelineKey(pipelineKey);
+      // default isHidden to true for new objects if not specified
+      runtimeMetadata.setHidden(isHidden == null || isHidden);
+    } else {
+      // row already existed in db; if caller did not provide value for isHidden, use existing value
+      runtimeMetadata.setHidden(isHidden == null ? runtimeMetadata.isHidden() : isHidden);
```
jsotobroad: can this be converted to 

```
runtimeMetadata.setPipelineKey(pipelineKey);
runtimeMetadata.setHidden(isHidden == null ? runtimeMetadata.isHidden() : isHidden);
```

I'm just not sure if we save a `null` for `isHidden` if the database will set it to true by default because of the `defaultValue` on the tables column (i feel like it should)

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
jsotobroad: ah i c.  I saw where we use in this in the ToolConfig class.  

is the builder used so we dont have to pass in teh `null`s?  I think a potiential downside is that if we add a new field that we want to be filled out when creating one of these it wont be immediately obvious where all the places we need to change are or enforce that a value is passed to that field.  We should definitely keep it since its convenient but just wanted to call out possible downsides.

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
jsotobroad: oh wait `pipelines.name` is already lower case in the current database table.  Why is necessary to have the `lower()` here?

service/src/main/java/bio/terra/pipelines/common/utils/PipelinesEnum.java
```diff
@@ -1,16 +1,31 @@
...
 public enum PipelinesEnum {
-  ARRAY_IMPUTATION("array_imputation"),
-  LOW_PASS_IMPUTATION("low_pass_imputation");
+  ARRAY_IMPUTATION("array_imputation", "arrayImputation"),
+  LOW_PASS_IMPUTATION("low_pass_imputation", "lowPassImputation");
+
+  @Getter private final String lowerCaseValue;
+  @Getter private final String configKeyValue;
 
-  private final String value;
```
jsotobroad, line 7: ahhhh `.getValue()` was already lower case.  you just renamed it to be more explicit. that makes sense.

service/src/main/java/bio/terra/pipelines/db/repositories/PipelineRuntimeMetadataRepository.java
jsotobroad: would it be worth keeping name and version in this table so we dont have to use `StartingWith` in this function?  I think its worth the extra columns to have direct comparisons in general even though this should never really be an issue

service/src/main/java/bio/terra/pipelines/common/utils/PipelineRunFilterSpecification.java
```diff
@@ -99,13 +96,10 @@ private static Predicate validateAndBuildJobIdPredicate(
...
   private static Predicate validateAndBuildPipelineNamePredicate(
-      String value,
-      Root<PipelineRun> root,
-      CriteriaQuery<?> query,
-      CriteriaBuilder criteriaBuilder) {
+      String value, Root<PipelineRun> root, CriteriaBuilder criteriaBuilder) {
     PipelinesEnum pipelineName;
     try {
-      pipelineName = PipelinesEnum.valueOf(value.toUpperCase());
+      pipelineName = PipelinesEnum.enumFromLowerCaseValue(value);
```
jsotobroad: I think you want to force value to be lower case here as well as this value comes from the user

service/src/main/java/bio/terra/pipelines/common/utils/PipelineRunFilterSpecification.java
jsotobroad: also is it worth adding pipeline name to the table so it can be a direct comparision instead of `like`?  Unlike the pipeline metadata table this table may actually get large enough for it to matter performance wise

service/src/test/java/bio/terra/pipelines/configuration/internal/PipelineConfigurationsTest.java
```diff
@@ -1,133 +1,645 @@
...
+    Set<String> inputNames = new HashSet<>();
+    for (PipelineConfigurations.PipelineInputDefinitionConfiguration input : inputs) {
+      if (input == null) {
+        violations.add("Input definition cannot be null for pipeline '%s'".formatted(pipelineKey));
+        continue;
+      }
+
+      String inputName = requireText(input.getName(), "inputs.name", pipelineKey, violations);
+      requireText(input.getWdlVariableName(), "inputs.wdlVariableName", pipelineKey, violations);
+      requireNonNull(input.getType(), "inputs.type", pipelineKey, violations);
```
jsotobroad, line 478: ah nice.  so the spring configuration magic will do the "is this the correct type" for us and this will make sure required fields are required, awesome!

service/src/test/java/bio/terra/pipelines/controller/PipelinesApiControllerTest.java
```diff
@@ -65,7 +65,8 @@ void beforeEach() {
     when(pipelinesServiceMock.getPipeline(any(PipelinesEnum.class), anyInt(), anyBoolean()))
         .thenReturn(getTestPipeline());
     when(quotasServiceMock.getPipelineQuota(any(PipelinesEnum.class)))
-        .thenReturn(TestUtils.TEST_PIPELINE_QUOTA_1);
```
jsotobroad, line 68: did we get rid of `TestUtils.TEST_PIPELINE_QUOTA_1`?

service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
```diff
@@ -18,15 +18,15 @@
...
+  @Autowired PipelineConfigurations pipelineConfigurations;
 
   @Test
   void getPipelineQuota() {
     PipelineQuota pipelineQuota = quotasService.getPipelineQuota(PipelinesEnum.ARRAY_IMPUTATION);
 
     assertEquals(PipelinesEnum.ARRAY_IMPUTATION, pipelineQuota.getPipelineName());
-    assertEquals(2500, pipelineQuota.getDefaultQuota());
-    assertEquals(500, pipelineQuota.getMinQuotaConsumed());
+    assertEquals(100, pipelineQuota.getDefaultQuota());
```
jsotobroad: any reason we couldnt have kept the test numbers the same so the other test related quota changes werent necessary?

jsotobroad (commented):
nice great progress!!

service/src/main/java/bio/terra/pipelines/db/repositories/PipelineRuntimeMetadataRepository.java
mmorgantaylor: i'm just adding pipelineName, since we can still get latest by ordering by key

service/src/main/java/bio/terra/pipelines/model/BasePipelineVariableDefinition.java
mmorgantaylor: yeah, that's the tradeoff. we hope to catch all the cases where we've forgotten something in tests.

service/src/main/java/bio/terra/pipelines/model/Pipeline.java
mmorgantaylor: no, i thought we landed on the intermediate solution of keeping Pipeline interface as-is but removing the PipelineDefinition and PipelineDefinitionProvider classes

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: lol there isn't one, thanks for catching this

mmorgantaylor: so they are the same, but I still want to chat about this

service/src/main/java/bio/terra/pipelines/service/PipelinesService.java
mmorgantaylor: ooh that did work. nice!

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
mmorgantaylor: you're right, not necessary here

service/src/main/java/bio/terra/pipelines/service/PipelineInputsOutputsService.java
mmorgantaylor: fun, the e2e tests i'm putting together now do, in fact, fail on local manifest upload. this'll fix those!

service/src/main/resources/db/changesets/20260527_add_pipeline_key_and_runtime_metadata.yaml
```diff
@@ -0,0 +1,138 @@
...
+            remarks: Runtime-managed metadata keyed by canonical pipeline_key.
+            columns:
+              - column:
+                  name: pipeline_key
+                  type: text
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: pipeline_name
```
jsotobroad, line 45: can we make this non nullable

mmorgantaylor: great minds

jsotobroad (approved):
thank you for the walkthroughs and all the changes!!

MatthewBemis (approved):

PR #407, opened 2026-06-10
[TSPS-929] [Minor version bump] New v2 /quotas admin API to accept email address

salonishah11:
### Description 

This PR adds v2 of both GET and PATCH /quotas endpoint that accept an email address to fetch or update quota for a user. The backend continues to use and store the user ID obtained from SAM in the quotas table to maintain backward compatibility and to avoid storing email addresses in our database. If the user exists in Terra and they don't have a quota row in database, the service will create a new quota row and return either the default quota (for GET) or new quota limit (for PATCH) for the pipeline instead of returning an error.

Related PR that will update tests to use the new v2 endpoints - https://github.com/broadinstitute/terra-github-workflows/pull/259

Jira ticket that will remove v1 of these endpoints in future - https://broadworkbench.atlassian.net/browse/TSPS-970

### Testing

Local testing

**GET endpoint**
User already had a record in TSPS database
```
{
  "userEmail": "groot.testerson@gmail.com",
  "userId": "101174847223453387077",
  "pipelineName": "array_imputation",
  "quotaLimit": 15000,
  "quotaConsumed": 4500
}
```

User did not have a record in TSPS database
```
{
  "userEmail": "sshah.testerson@gmail.com",
  "userId": "115283646847622313400",
  "pipelineName": "array_imputation",
  "quotaLimit": 2500,          <---- default free quota for array_imputation
  "quotaConsumed": 0
}
```
New record added for the user
Screenshot 2026-06-10 at 9 49 17 AM

User doesn't exist in SAM
```
{
  "message": "User with email 'my.testerson@gmail.com' not found in SAM",
  "statusCode": 404,
  "causes": []
}
```

User calling the admin endpoint is not an admin
```
{
  "message": "User groot.testerson@gmail.com is not authorized to perform admin action",
  "statusCode": 403,
  "causes": []
}
```

**PATCH endpoint**

Error checks are same as GET endpoint

Update called for a user that doesn't exist in database
Screenshot 2026-06-10 at 9 53 06 AM

```
{
  "userEmail": "sshah.testerson@gmail.com",
  "userId": "115283646847622313400",
  "pipelineName": "array_imputation",
  "quotaLimit": 5000,
  "quotaConsumed": 0
}
```

New record created in database with new limit
Screenshot 2026-06-10 at 9 54 02 AM

- [x] Planned non patch version bump (if applicable)

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -54,6 +54,7 @@ paths:
       - $ref: '#/components/parameters/PipelineName'
       - $ref: '#/components/parameters/UserId'
     get:
+      deprecated: true
       summary: Get quota for a given pipeline and user.
       tags: [ admin ]
       operationId: getQuotaForPipelineAndUser
@@ -71,6 +72,7 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
     patch:
+      deprecated: true
       summary: Update quota limit for a given pipeline and user.
       tags: [ admin ]
       operationId: updateQuotaLimitForPipelineAndUser
@@ -87,6 +89,51 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/AdminQuota'
+        400:
+          $ref: '#/components/responses/BadRequest'
+        403:
+          $ref: '#/components/responses/PermissionDenied'
+        500:
+          $ref: '#/components/responses/ServerError'
+
+  /api/admin/v2/quotas/{pipelineName}/{userEmail}:
+    parameters:
+      - $ref: '#/components/parameters/PipelineName'
+      - $ref: '#/components/parameters/UserEmail'
+    get:
+      summary: Get quota for a given pipeline and user identified by email
+      tags: [ admin ]
+      operationId: getQuotaForPipelineAndUserV2
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdminQuotaV2'
+        400:
+          $ref: '#/components/responses/BadRequest'
+        403:
+          $ref: '#/components/responses/PermissionDenied'
+        500:
+          $ref: '#/components/responses/ServerError'
+    patch:
+      summary: Update quota limit for a given pipeline and user identified by email
+      tags: [ admin ]
+      operationId: updateQuotaLimitForPipelineAndUserV2
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UpdateQuotaLimitRequestBody'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdminQuotaV2'
         400:
           $ref: '#/components/responses/BadRequest'
         403:
@@ -617,6 +664,15 @@ components:
       schema:
         type: string
 
+    UserEmail:
+      name: userEmail
+      in: path
+      description: The email address of the Terra user
+      required: true
+      schema:
+        type: string
+        format: email
+
     WorkspaceId:
       name: workspaceId
       in: path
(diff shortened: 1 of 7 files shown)


common/openapi.yml
jsotobroad, line 885: should this be deprecated now?

service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
```diff
@@ -189,6 +190,60 @@ void updateQuotaConsumedGreaterThanQuota() {
         () -> quotasService.updateQuotaConsumed(userQuota, newQuotaConsumed));
   }
 
+  @Test
+  void createQuotaForUserAndPipeline() {
```
jsotobroad: just curious why these tests were added. I assumed this would already exist

salonishah11: This tests the new method `createQuotaForUserAndPipeline` that was added in QuotasService class. This is different from the existing `getOrCreateQuotaForUserAndPipeline` method

common/openapi.yml
salonishah11: Oh good point. I will add the deprecated tag here also

salonishah11: this is ✅

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -157,6 +170,90 @@ public ResponseEntity<ApiAdminQuota> updateQuotaLimitForPipelineAndUser(
...
+    }
+
+    // this is used for email notification
+    int originalQuotaLimit = 0;
+
+    int newQuotaLimit = body.getQuotaLimit();
+    UserQuota newUserQuota;
+
+    Optional<UserQuota> currentUserQuotaOptional =
+        quotasService.getQuotaForUserAndPipeline(userId, validatedPipelineName);
```
jsotobroad: couldnt you use the getOrCreateQuotaForUserAndPipeline function here which will always return a UserQuota and then update the quota from there? Then you wouldnt need to add the function in QuotasService.

salonishah11: Yes we can. But that would create additional database call. And we would be adding a row just to update it in the next statement which seemed like a weird pattern to me.

jsotobroad: so theoretically there will always be a row before you make this api call cuz the admin has to make sure they know what the users final quota should be (current quota + quota purchased).

I would take an extra db call in an admin api for simplified logic in the controller.  If you really want to only  do one database call then i would move this logic into `QuotasService`

salonishah11: I found creating a row and updating it in the next call a bit of weird pattern. I like the idea of moving the logic to QuotasService class.

One thing I was thinking, instead of passing the new `quotaLimit` as a field in request body, why not just pass the the quota that should be added in the request body and the service can handle the logic? Wouldn't this make the PATCH endpoint easier?

jsotobroad: this would make it so we couldnt reduce a users quota and i think thats something we want to be able to do.

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
```diff
@@ -157,6 +170,90 @@ public ResponseEntity<ApiAdminQuota> updateQuotaLimitForPipelineAndUser(
...
+        PipelineApiUtils.validatePipelineName(pipelineName, logger);
+
+    // get userId from Sam using the email address and throw not found error
+    // if the user doesn't exist in Sam
+    String userId;
+    try {
+      userId = samService.getUserIdFromEmail(authedUser, userEmail);
+    } catch (bio.terra.common.exception.NotFoundException e) {
+      throw new NotFoundException(
+          String.format("User with email '%s' not found in SAM", userEmail), e);
```
jsotobroad, line 190: what is the default error message from the `bio.terra.common.exception.NotFoundException` that is being caught here?

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
salonishah11: Oh yeah, that's a good point.

service/src/main/java/bio/terra/pipelines/app/controller/AdminApiController.java
salonishah11: Sam actually doesn't return any error message. It just returns 404. So this catches it and re-throws it with better error message

jsotobroad: gotcha, thanks!

jsotobroad (approved):

MatthewBemis (approved):
awesome! this will be way more ergonomic

PR #419, opened 2026-06-22
[TSPS-943] Add validationRegex to pipeline input definitions

MatthewBemis:
### Description 

Adds a new optional validationRegex property to user input strings.

The first implementation of this is to use it for outputBasename in array_imputation and low_pass_imputation.

If a user submits bad inputs:

```
{
  "message": "Problem with pipelineInputs: outputBasename must only contain alphanumeric characters, dashes, and underscores, and be at most 255 characters",
  "statusCode": 400,
  "causes": []
}
```

Returned in pipeline details, for potential future use in the UI and/or CLI:

```
...
    {
      "name": "outputBasename",
      "displayName": "output basename",
      "description": "The prefix for all of the outputs' filenames. May only contain alphanumeric characters, dashes, and underscores",
      "type": "STRING",
      "isRequired": true,
      "validationRegex": "^[a-zA-Z0-9_-]{1,255}$",
      "validationRegexExplanation": "outputBasename must only contain alphanumeric characters, dashes, and underscores, and be at most 255 characters"
    },
...
```

### Jira Ticket

https://broadworkbench.atlassian.net/browse/TSPS-943

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
@@ -124,7 +124,8 @@ public class PipelinesApiController implements PipelinesApi {
                         .minValue(input.getMinValue())
                         .maxValue(input.getMaxValue())
                         .fileSuffix(input.getFileSuffix())
-                        .validationRegex(input.getValidationRegex()))
+                        .validationRegex(input.getValidationRegex())
+                        .validationRegexExplanation(input.getValidationRegexExplanation()))
             .toList());
     ApiPipelineOutputDefinitions outputs = new ApiPipelineOutputDefinitions();
     outputs.addAll(
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
@@ -124,8 +124,9 @@ public class PipelinesApiController implements PipelinesApi {
                         .minValue(input.getMinValue())
                         .maxValue(input.getMaxValue())
                         .fileSuffix(input.getFileSuffix())
-                        .validationRegex(input.getValidationRegex())
-                        .validationRegexExplanation(input.getValidationRegexExplanation()))
+                        .validationRegex(input.getType().getEffectiveValidationRegex(input))
+                        .validationRegexExplanation(
+                            input.getType().getEffectiveValidationExplanation(input)))
             .toList());
     ApiPipelineOutputDefinitions outputs = new ApiPipelineOutputDefinitions();
     outputs.addAll(
--- a/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
+++ b/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
@@ -123,7 +123,8 @@ public class PipelinesApiController implements PipelinesApi {
                         .defaultValue(input.getDefaultValue())
                         .minValue(input.getMinValue())
                         .maxValue(input.getMaxValue())
-                        .fileSuffix(input.getFileSuffix()))
+                        .fileSuffix(input.getFileSuffix())
+                        .validationRegex(input.getValidationRegex()))
             .toList());
     ApiPipelineOutputDefinitions outputs = new ApiPipelineOutputDefinitions();
     outputs.addAll(
--- a/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
+++ b/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
@@ -39,10 +39,14 @@ public enum PipelineVariableTypesEnum {
       if (effectivePattern.matcher(castValue).matches()) {
         return null;
       }
-      return hasCustomRegex
-          ? "%s must match the pattern %s".formatted(fieldName, validationRegex)
-          : "%s must only contain alphanumeric characters or the following symbols: %s"
-              .formatted(fieldName, DEFAULT_VALID_STRING_PATTERN_SYMBOLS);
+      if (hasCustomRegex) {
+        String explanation = pipelineInputDefinition.getValidationRegexExplanation();
+        return explanation != null
+            ? "%s %s".formatted(fieldName, explanation)
+            : "%s must match the pattern %s".formatted(fieldName, validationRegex);
+      }
+      return "%s must only contain alphanumeric characters or the following symbols: %s"
+          .formatted(fieldName, DEFAULT_VALID_STRING_PATTERN_SYMBOLS);
     }
   },
   INTEGER {
--- a/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
+++ b/service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
@@ -28,18 +28,21 @@ public enum PipelineVariableTypesEnum {
       if (castValue == null) {
         return "%s must be a string".formatted(fieldName);
       }
+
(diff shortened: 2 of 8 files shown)


service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -28,11 +28,21 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
       if (castValue == null) {
         return "%s must be a string".formatted(fieldName);
       }
-      if (!VALID_STRING_PATTERN.matcher(castValue).matches()) {
-        return "%s must only contain alphanumeric characters or the following symbols: %s"
-            .formatted(fieldName, VALID_STRING_PATTERN_SYMBOLS);
+
+      // validationRegex, when present, fully replaces the default VALID_STRING_PATTERN.
```
MatthewBemis, line 32: we could get rid of the default pattern if we added validation regexes to _every_ string input definition. but i like having the default as a fallback and only specifying a more specific pattern as-needed.

mmorgantaylor: i agree!

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
```diff
@@ -123,7 +123,8 @@ static ApiPipelineWithDetails pipelineWithDetailsToApi(Pipeline pipelineInfo) {
                         .defaultValue(input.getDefaultValue())
                         .minValue(input.getMinValue())
                         .maxValue(input.getMaxValue())
-                        .fileSuffix(input.getFileSuffix()))
+                        .fileSuffix(input.getFileSuffix())
+                        .validationRegex(input.getValidationRegex()))
```
mmorgantaylor: just so I understand, this will only be populated if there's a custom one in the config, right? i.e. we won't return the default string regex if there's no custom one?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -28,11 +28,21 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
...
+      boolean hasCustomRegex = validationRegex != null;
+      Pattern effectivePattern =
+          hasCustomRegex ? Pattern.compile(validationRegex) : DEFAULT_VALID_STRING_PATTERN;
+
+      if (effectivePattern.matcher(castValue).matches()) {
+        return null;
       }
-      return null;
+      return hasCustomRegex
+          ? "%s must match the pattern %s".formatted(fieldName, validationRegex)
```
mmorgantaylor: I worry that most of our users won't understand this regex notation. one option could be to add another field to the pipeline input definition that's like `validationRegexExplanation` or something that writes it out in plain text that users can understand. we could have the config validator ensure that if there's a `validationRegex` field, the `validationRegexExplanation` field is also populated.

mmorgantaylor (commented):
very cool! a couple questions

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
MatthewBemis: This crossed my mind too but I convinced myself it was okay because only API users would see this and they're _hopefully_ savvy enough to figure it out.

That said, I really like your `validationRegexExplanation ` idea. I can add that in. That may end up simplifying some things in the UI too.

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
MatthewBemis: Yes- that's correct

jsotobroad (commented):
is the work for capping the length of all strings gonna be in another pr or should it be added to this one?

MatthewBemis:
@jsotobroad it's in this PR already- the default regex for string validation was changed here: https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/419/changes#diff-0747cf19c4c924c722337946da42642a58f31ccb0bc90768a0d2e254c5fec81fR282

jsotobroad:
> @jsotobroad it's in this PR already- the default regex for string validation was changed here: https://github.com/DataBiosphere/terra-scientific-pipelines-service/pull/419/changes#diff-0747cf19c4c924c722337946da42642a58f31ccb0bc90768a0d2e254c5fec81fR282

hmmm is that the best plce to do it since now if we have a custom regex we need to remember to add that part to the regex?  Is there any reason to not just add it to the "String" type validation?

MatthewBemis:
@jsotobroad i think so. i consider length requirements to be a part of input validation so having it bundled together with the character validation in one single regex seems cleanest to me. and if we ever have an input string that requires a shorter max length than 255, this would keep things consistent regarding how we define max length

jsotobroad:
> @jsotobroad i think so. i consider length requirements to be a part of input validation so having it bundled together with the character validation in one single regex seems cleanest to me. and if we ever have an input string that requires a shorter max length than 255, this would keep things consistent regarding how we define max length

gotcha, fair enough!

service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java
jsotobroad: why wouldnt we want to return the custom regex for string inputs that dont have a custom one?

MatthewBemis: I don't have a good reason- we should do it

jsotobroad (approved):
nice!

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -28,11 +28,35 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
...
+    @Override
+    public String getEffectiveValidationRegex(PipelineInputDefinition def) {
+      return def.getValidationRegex() != null
+          ? def.getValidationRegex()
+          : DEFAULT_VALID_STRING_REGEX;
+    }
+
+    @Override
+    public String getEffectiveValidationExplanation(PipelineInputDefinition def) {
+      if (def.getValidationRegexExplanation() != null) {
```
mmorgantaylor, line 53: we could alternatively add a PipelineConfigurationsTest test to ensure that if there's a custom validation regex, there must be a validation regex explanation, and then we can simplify this logic a bit, right?

service/src/main/java/bio/terra/pipelines/common/utils/PipelineVariableTypesEnum.java
```diff
@@ -265,12 +289,34 @@ public String validate(PipelineInputDefinition pipelineInputDefinition, Object v
    */
   public abstract String validate(PipelineInputDefinition pipelineInputDefinition, Object value);
 
+  /**
+   * Returns the effective validation regex string for this input type. Returns null for types that
+   * do not have a regex-based validation concept. STRING overrides this to return the configured
+   * regex, or {@link #DEFAULT_VALID_STRING_REGEX} when none is configured.
+   */
+  public String getEffectiveValidationRegex(PipelineInputDefinition pipelineInputDefinition) {
```
mmorgantaylor, line 297: mark this sonar finding as false positive, since it gets overridden in STRING? (same for below)

mmorgantaylor (commented):
a suggestion but if you don't like it that's ok

mmorgantaylor (approved):
giving you a thumb so you're not blocked when i'm out; up to you if you want to add more validation to PipelineConfigurationsTest in order to simplify the getEffectiveValidationExplanation logic. might not be worth it, and I'm fine with what you have.

PR #420, opened 2026-06-23
TSPS-999 address duplicate id issues found in e2e tests

jsotobroad:
### Description 

We have some live e2e tests failing with "duplicate ids".  It tuns out this was caused partly by having the stairway job submit code wrapped in a transaction.  Because its a different database we cant actually roll back what its doing so on retry, it would retry the submit which would return the duplicate id issue.  

Database issue causing the retry
```
2026-06-23 13:10:47.249 UTC [1654922]: [4-1] db=teaspoons,user=teaspoons ERROR:  could not serialize access due to read/write dependencies among transactions
```
We think with these changes, that erorr will be retried until teaspoons succeeds and then the flight will be kicked off in stairway

Bee e2e test from this branch:
https://github.com/DataBiosphere/terra-scientific-pipelines-service/actions/runs/28051465384

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-999

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
@@ -36,6 +36,7 @@ import org.hibernate.exception.ConstraintViolationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
```diff
@@ -162,10 +176,47 @@ public Map<String, Map<String, String>> preparePipelineRunV2(
...
+    StartPipelineRunDbResult dbResult =
+        self.startPipelineRunInTransaction(pipeline, jobId, authedUser);
+
+    // Step 2: Submit the Stairway flight only after the DB transaction has committed.
+    try {
+      dbResult.jobBuilder().submit();
+      logger.info("Started {} startedPipelineRun with jobId {}", pipelineName, jobId);
+      return dbResult.pipelineRun();
+    } catch (Exception e) {
+      markPipelineRunFailed(jobId, userId);
```
mmorgantaylor, line 202: if a fatal error happens in `self.startPipelineRunInTransaction` will we ever mark the pipeline run failed? can we make sure we have a test for that scenario? (the new test is for errors with stairway submit)

mmorgantaylor (commented):
please make a ticket for this work (required by SDLC)

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
jsotobroad: ahhh nice catch!  Might not get to it today but will update the code/tests

mmorgantaylor (approved):
thank you for figuring this out!

salonishah11 (approved):

service/src/main/java/bio/terra/pipelines/service/PipelineRunsService.java
jsotobroad: after talkign in post, we think its ok to not mark the pipeline as failed if something fails in this function

PR #421, opened 2026-06-25
TSPS-991 migrate Java codegen from swagger-codegen to openapi-generator  [MAJOR]

mmorgantaylor:
### Description 

Migrates the three Java codegen sites (rawls-client, client, service) from the dormant `swagger-codegen-cli` to `openapi-generator`, matching python-client. Applied on an isolated plugin classpath (version centralized in settings.gradle) so dependabot can keep codegen current. 

swagger-codegen + the hidetake plugin are fully removed.

NOTE: published Java client is source-breaking (but wire-compatible): 
- 6 container-typedef classes inlined to Map/List (e.g. `ApiPipelineRunOutputs` becomes just `Map<String, Object>`)
- `SystemStatusSystems` is now `SystemStatusSystemsValue` 
- This is what prompts the major bump (for Java client users, of which we have none, but it's the right thing to do)

### Jira Ticket
https://broadworkbench.atlassian.net/browse/TSPS-991


service/src/test/java/bio/terra/pipelines/dependencies/rawls/RawlsClientTest.java
```diff
@@ -27,7 +27,7 @@ void testWorkspaceAuthorizedClient() {
     apiClient = rawlsClient.getApiClient(authToken);
     for (Authentication auth : apiClient.getAuthentications().values()) {
       if (auth instanceof OAuth) {
-        String actualToken = ((OAuth) auth).getAccessToken();
+        String actualToken = ((OAuth) auth).getAccessToken().getAccessToken();
```
mmorgantaylor, line 30: this is a difference due to a different OAuth.java used by swagger-codegen (old) and openapi-generator (new). 

note this only applies to rawls and published client builds.

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -60,22 +59,6 @@ public WorkspaceDetails getWorkspaceDetails(
                 .getWorkspace());
   }
 
-  public String getWorkspaceBucketName(WorkspaceDetails workspaceDetails) {
```
mmorgantaylor, line 63: openapi-generator now enforces that bucketName and googleProject are not null, so we don't need this error handling or these methods.

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -443,11 +443,11 @@ public ResponseEntity<ApiGetPipelineRunsResponseV2> getAllPipelineRunsV2(
   private ApiAsyncPipelineRunResponseV2 pipelineRunToApiV2(
       PipelineRun pipelineRun,
       String pipelineKey,
-      Function<PipelineRun, ApiPipelineRunOutputs> fetchOutputsFunction) {
+      Function<PipelineRun, Map<String, Object>> fetchOutputsFunction) {
```
mmorgantaylor, line 446: one big difference in this migration is that we don't have these obscuring class names for API objects that are just Maps. I think this is much clearer.

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
```diff
@@ -443,11 +443,11 @@ public ResponseEntity<ApiGetPipelineRunsResponseV2> getAllPipelineRunsV2(
...
     ApiAsyncPipelineRunResponseV2 response = new ApiAsyncPipelineRunResponseV2();
 
-    ApiPipelineUserProvidedInputs userProvidedInputs = new ApiPipelineUserProvidedInputs();
-    userProvidedInputs.putAll(pipelineInputsOutputsService.retrieveUserProvidedInputs(pipelineRun));
+    Map<String, Object> userProvidedInputs =
```
mmorgantaylor, line 449: a benefit of the above change in object/class name is that we can instantiate new instances much more cleanly

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
```diff
@@ -86,7 +69,7 @@ public SubmissionReport submitWorkflow(
         () ->
             rawlsClient
                 .getSubmissionsApi(accessToken)
-                .createSubmission(submissionRequest, workspaceNamespace, workspaceName));
+                .createSubmission(workspaceNamespace, workspaceName, submissionRequest));
```
mmorgantaylor, line 86: this change in input variable order is just cause the new generator does things differently. easy enough to update our code.

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
jsotobroad: I think you lose a bit useful context by just creating/passing around generic maps but if this is what it does then thats fine

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
jsotobroad: what is the thing that causes this to be enforced now? like did the rawls yml have this as required but the previous library want abiding by it and this new library is?

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
jsotobroad: do we know what causes this in the new generator?  seems weird that this would be a different order from the same source

service/src/main/java/bio/terra/pipelines/service/BaseStatusService.java
```diff
@@ -1,7 +1,7 @@
 package bio.terra.pipelines.service;
 
 import bio.terra.pipelines.app.configuration.internal.StatusCheckConfiguration;
-import bio.terra.pipelines.generated.model.ApiSystemStatusSystems;
+import bio.terra.pipelines.generated.model.ApiSystemStatusSystemsValue;
```
jsotobroad, line 4: any intuition as to why the added `Value` ?

jsotobroad (commented):
just random annoying questions :)

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
mmorgantaylor: it's just a generator convention difference in how each tool orders the generated method's parameters. swagger-codegen put the request body before the path parameters, and openapi-generator puts the path parameters first and then request body.

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
mmorgantaylor: so yeah this is fun. in the Rawls openApi spec, `bucketName` and `googleProject` are listed under the `required:` fields. swagger-codegen didn't do anything with this; openapi-generator sees this and adds a @Nonnull annotation. so it's not quite that we're enforcing not null, but based on their openApi spec, we expect them to be there because they are marked required.

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
mmorgantaylor: I can see that, but I hated the implications this had for populating these objects. they're really just a map. the variable names we use should make it clear what they are (and happy to update them if any aren't clear)

service/src/main/java/bio/terra/pipelines/service/BaseStatusService.java
mmorgantaylor: this is again a generator naming convention difference.                                                                       

in `common/openapi.yml`, `SystemStatus` has a systems field that's a map (`additionalProperties`) whose value is an inline object schema (the per-subsystem status: { ok, messages }). That inline value-type has no name of its own in the spec, so the generator has to synthesize one:                                                                                                                                                                                                               
  - swagger-codegen named it after the parent + property: SystemStatusSystems                                                                                                                                      
  - openapi-generator appends Value to mark it as the value type of a map/additionalProperties schema: SystemStatusSystemsValue

🤷

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
mmorgantaylor: ok so since the annotation doesn't enforce those fields to be not null, I added a null check on those when we get them from Rawls - a better flow, so this will fail if those are null when the admin is setting the workspace, not later when these values are being retrieved.

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
jsotobroad: gotcha, thanks!

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
jsotobroad: def dont need to update them back to objects

service/src/main/java/bio/terra/pipelines/dependencies/rawls/RawlsService.java
jsotobroad: huh its a little lame that the neither generator enforces non null if the openApi spec yml has them as `required`.  the check at fetching makes sense to me

service/src/main/java/bio/terra/pipelines/service/BaseStatusService.java
jsotobroad: 👍  thanks!

service/src/main/java/bio/terra/pipelines/app/controller/PipelineRunsApiController.java
mmorgantaylor: def wasn't offering to do that 😀

service/src/test/java/bio/terra/pipelines/dependencies/rawls/RawlsClientTest.java
salonishah11: `.getAccessToken().getAccessToken()` is weird but 🤷‍♀️

salonishah11 (approved):

MatthewBemis (approved):

PR #429, opened 2026-07-08
[TSPS-1028] Fix race condition when creating quota row for new user

salonishah11:
### Description 

This PR adds a try-catch in `getOrCreateQuotaForUserAndPipeline` to catch the DataIntegrityViolationException when multiple threads try to fetch and insert a row for a user whose quota doesn't exist in the table. 

Alternative solutions considered:
- UPSERT: Native SQL UPSERT would require custom queries and bypass Spring Data JPA and add complexity
- Table Locking: Pessimistic locking would require `@Transactional` context and hold locks during the entire transaction, significantly reducing throughput
- Row Locking: Cannot lock a row that doesn't exist yet

Try-catch approach leverages the existing unique constraint and works with the current repository pattern. It also assumes conflicts are rare (since this race condition only happens when multiple threads are called to fetch data for a new user) and only retries on actual collisions, providing better performance.

I have added information about why this race condition might be happening in ticket comment - https://broadworkbench.atlassian.net/browse/TSPS-1028?focusedCommentId=116091 

### DIFF
--- a/service/src/main/java/bio/terra/pipelines/service/QuotasService.java
+++ b/service/src/main/java/bio/terra/pipelines/service/QuotasService.java
@@ -52,6 +52,7 @@ public class QuotasService {
   public UserQuota getOrCreateQuotaForUserAndPipeline(String userId, PipelinesEnum pipelineName) {
     // try to get the user quota
     Optional<UserQuota> userQuota = getQuotaForUserAndPipeline(userId, pipelineName);
+
     // if the user quota is not found, grab the default pipeline quota and make a new row in user
     // quotas table
     if (userQuota.isEmpty()) {
@@ -59,13 +60,32 @@ public class QuotasService {
           "Couldn't find user quota for user {} and pipeline {}. Creating a new row",
           userId,
           pipelineName);
-      PipelineQuota pipelineQuota = getPipelineQuota(pipelineName);
-      UserQuota newUserQuota = new UserQuota();
-      newUserQuota.setUserId(userId);
-      newUserQuota.setPipelineName(pipelineName);
-      newUserQuota.setQuota(pipelineQuota.getDefaultQuota());
-      userQuotasRepository.save(newUserQuota);
-      return newUserQuota;
+      try {
+        /* Note: This try/catch relies on this method NOT being wrapped in an outer @Transactional.
+         * Because there is no outer transaction, CrudRepository.save() commits and flushes immediately,
+         * allowing us to catch the constraint violation. If you add @Transactional to this service
+         * in the future, the flush will be delayed and the try/catch will fail.
+         */
+        PipelineQuota pipelineQuota = getPipelineQuota(pipelineName);
+        UserQuota newUserQuota = new UserQuota();
+        newUserQuota.setUserId(userId);
+        newUserQuota.setPipelineName(pipelineName);
+        newUserQuota.setQuota(pipelineQuota.getDefaultQuota());
+        return userQuotasRepository.save(newUserQuota);
+      } catch (org.springframework.dao.DataIntegrityViolationException e) {
+        // Handle race condition: another thread may have created the quota between our check and
+        // insert
+        logger.info(
+            "Race condition detected while creating user quota for user {} and pipeline {}. Fetching existing record.",
+            userId,
+            pipelineName);
+        return getQuotaForUserAndPipeline(userId, pipelineName)
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Quota should exist but was not found for user %s and pipeline %s."
+                            .formatted(userId, pipelineName)));
+      }
     }
     return userQuota.get();
   }
--- a/service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
+++ b/service/src/test/java/bio/terra/pipelines/service/QuotasServiceTest.java
@@ -370,10 +370,10 @@ class QuotasServiceTest extends BaseEmbeddedDbTest {
             IllegalStateException.class,
             () -> serviceWithMock.getOrCreateQuotaForUserAndPipeline(userId, pipeline));
 
-    // Verify exception message contains user and pipeline info
-    assertTrue(exception.getMessage().contains(userId));
-    assertTrue(exception.getMessage().contains(pipeline.name()));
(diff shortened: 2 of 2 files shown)


service/src/main/java/bio/terra/pipelines/service/QuotasService.java
```diff
@@ -52,20 +52,40 @@ public QuotaUnitsEnum getQuotaUnitsForPipeline(PipelinesEnum pipelineName) {
...
+        logger.info(
+            "Race condition detected while creating user quota for user {} and pipeline {}. Fetching existing record.",
+            userId,
+            pipelineName);
+        return getQuotaForUserAndPipeline(userId, pipelineName)
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Quota should exist but was not found for user %s and pipeline %s."
+                            .formatted(userId, pipelineName)));
```
salonishah11, line 87: Sonarcloud is complaining about missing code coverage for these lines. I am fine with ignoring the Sonarcloud issue but happy to add a test if needed.

I have added `orElseThrow` as a safety net which should most likely never be called since the record should exist if the `DataIntegrityViolationException` happens

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
```diff
@@ -52,20 +52,40 @@ public QuotaUnitsEnum getQuotaUnitsForPipeline(PipelinesEnum pipelineName) {
...
-      newUserQuota.setUserId(userId);
-      newUserQuota.setPipelineName(pipelineName);
-      newUserQuota.setQuota(pipelineQuota.getDefaultQuota());
-      userQuotasRepository.save(newUserQuota);
-      return newUserQuota;
+      try {
+        /* Note: This try/catch relies on this method NOT being wrapped in an outer @Transactional.
+         * Because there is no outer transaction, CrudRepository.save() commits and flushes immediately,
+         * allowing us to catch the constraint violation. If you add @Transactional to this service
+         * in the future, the flush will be delayed and the try/catch will fail.
```
jsotobroad, line 67: would adding a `@Transactional` cause this error to not happen anymore?  It seems like we think this would reduce throughput, do we have an idea of what the reduction would be?  Unless it is a significant reduction I would rather us fix the race condition with this rather than ignore it.

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
jsotobroad: hmm querying the database seems weird here. is it really necessary?

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
salonishah11: My understanding is that adding `@Transactional` to original code won't fix the issue since now both fetch and insert happen in a single commit which would just delay the DataViolation exception no?

One way to fix the race condition is using Upsert with custom SQL query. Did you have other suggestions?

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
salonishah11: What would be your suggestion instead of querying for the user quota?

jsotobroad: oh right this is supposed to return the quota hahaha, sorry!  this isnt _just_ inserting a row. nvm!!

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
jsotobroad: hmm but isnt this comment saying the try catch wont work if this is in a `@Transactional`.  So what does happen instead?

In your description you have
```
Table Locking: Pessimistic locking would require @Transactional context and hold locks during the entire transaction, significantly reducing throughput
```
What else would we need to add other than `@Transactional` for this to work?

salonishah11: My understanding (with AI's help) of `@Transactional` is that adding it would delay the commit after the method returns.  So the constraint violation happens during commit (outside the try-catch block), not during save(). The exception becomes a transaction rollback that propagates to the caller instead of being caught. So the exception is thrown outside method and the try-catch block never executes and so we would still see the DataViolation exception.

> What else would we need to add other than `@Transactional` for this to work?

I think to make `@transactional` work with table locking, we would need to use native SQL query and locks

salonishah11: I prefer the current approach because this race condition happens rarely and only for for new users. Most of the time fetch is what is used in `getOrCreateQuotaForUserAndPipeline` and so locking the entire table for these rare occurrences seems overkill. 

Below is what AI's suggested approach is for table locking + transactional --

#### Table Locking (Pessimistic Concurrency)

##### Required Changes:

1. Add EntityManager dependency
2. Add @Transactional annotation
3. Use native SQL to lock table
4. Change repository inheritance (optional, but recommended)

#### PostgreSQL Advisory Locks

**Advisory locks** are application-level locks that don't lock actual table rows

##### Required Changes:

1. Add EntityManager dependency
2. Add @Transactional annotation
3. Import java.util.Objects for hash generation
4. Use native SQL to acquire advisory lock

jsotobroad: gotcha thank you for the in depth explanation!  this looks good to me

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
jsotobroad: would adding a test to cover this be difficult?  if not i would say lets just add it

jsotobroad (approved):
thank you for all the explanations!  Just one question about the sonarqube test but looks great

service/src/main/java/bio/terra/pipelines/service/QuotasService.java
salonishah11: Let me try that.

salonishah11: AI was able to do this easily and hence I have added a unit test.

MatthewBemis (approved):
thanks!

